PR #3, opened 2023-04-17
Use anchors per draw command

fenjalien:
This adds a field called `name` on all draw commands with anchors (`line`, `rect` and  `circle`) which is `none` by default but should be set to a string. `ctx` then only stores anchors if an element both has anchors and a non-none name.

This does break `node` and `anchor`


johannes-wolf:
LGTM. I will fix-up the errors with `node(..)`

PR #9, opened 2023-04-17
Fix control points for `line` command

fenjalien:
You currently can't use Bezier curves with the `line` command even though its drawn by Typst's `path`. 

This PR solves this by checking if the first element of a vector is an array whenever an operation is to be performed on it, if so the operation is applied to each vector in the array. `apply-transform` and the final translation are also not applied to control points, this fixes the fact that control points are supposed to be relative but any scale or rotation would break them.

As a side note @johannes-wolf you may want to switch to using squash commits instead of merge.

### DIFF
--- a/canvas.typ
+++ b/canvas.typ
@@ -14,7 +14,8 @@
   // Apply all transformation matrices `queue` in order
   // on `vec`.
   let apply-transform(queue, vec) = {
-    if type(vec) == "array" and type(vec.first()) != "float" {
+    if type(vec.first()) == "array" {
+      // the vector has control points, don't transform them
       vec.at(0) = apply-transform(queue, vec.first())
     } else {
     for (k, m) in queue {
@@ -86,7 +87,7 @@
   let bounding-box(pts, init: none) = {
     let bounds = init
     for (i, pt) in pts.enumerate() {
-      // check if point has control points
+      // check if point has control points, only use the point itself not the control points
       if type(pt.first()) == "array" {
         pt = pt.first()
       }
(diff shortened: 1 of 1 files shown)


canvas.typ
```diff
@@ -250,12 +265,21 @@
       place(dx: pt.at(0), dy: pt.at(1), self.content)
     }
   );
-  
   box(width: width, height: height, fill: fill, {
+
+    let vec-to-pt(vec) = {
+      // the vector has control points
+      if type(vec.first()) == "array" {
+        return vec.map(slice_vec)
```
johannes-wolf, line 273: This function is missing from the PR.

johannes-wolf (changes requested):
One problem with (typst native) curves is, that we are not able to determine their bounding box to correctly resize the canvas.

I also think, the control points have to be transformed along with their path vertices.

fenjalien:
To give an example of why control points shouldn't be translated along with the vertex:

The path
```typst
#path(
  ((0pt,-20pt), (0pt,0pt), (20pt, 0pt)),
  (20pt,0pt),
  ((0pt,20pt), (20pt,0pt), (0pt, 0pt)),
)
```
produces
image

to translate the path `10pt` to the right, translate only the vertices
```typst
#path(
  ((10pt,-20pt), (0pt,0pt), (20pt, 0pt)),
  (30pt,0pt),
  ((10pt,20pt), (20pt,0pt), (0pt, 0pt)),
)
```
which produces the same image as above. If you also translate the control points
```typst
#path(
  ((10pt,-20pt), (10pt,0pt), (30pt, 0pt)),
  (30pt,0pt),
  ((10pt,20pt), (30pt,0pt), (10pt, 0pt)),
)
```
produces 
image
which is incorrect.

canvas.typ
fenjalien: oops missed that

fenjalien:
Actually scaling and rotation transforms might be okay, it would just be translations?

johannes-wolf:
I would like to close this in favour of #10 
Reasons:
- With a custom impl. we get correct sizing behaviour of the canvas
- Passing control-points in arrays insides vectors feels hacky
- We will be able to calculate a given bezier point for placing objects at paths

johannes-wolf:
I do use merge commits because they make for a nice history. Cleaning and rebasing should be done to PRs before merging.

PR #22, opened 2023-04-22
Fix anchor transformations for good

fenjalien:
Also fixed rotations and flipped translating in the y direction

### DIFF
--- a/draw.typ
+++ b/draw.typ
@@ -44,7 +44,7 @@
         ctx.transform.undo.push(matrix.transform-rotate-xyz(-x, -y, -z))
       } else {
         ctx.transform.do.push(matrix.transform-rotate-z(angle))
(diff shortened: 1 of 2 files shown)


fenjalien:
looks like the shadow test is failing due to the y translation flip. Should the test source be changed or update the image? For the latter I can't as I'm on windows not linux.

draw.typ
```diff
@@ -44,7 +44,7 @@
         ctx.transform.undo.push(matrix.transform-rotate-xyz(-x, -y, -z))
       } else {
         ctx.transform.do.push(matrix.transform-rotate-z(angle))
-        ctx.transform.undo.push(matrix.transform-rotate-z(angle))
+        ctx.transform.undo.push(matrix.transform-rotate-z(-angle))
```
johannes-wolf, line 47: Ooops :D

PR #25, opened 2023-04-22
Add function coordinate syntax to `rotate` function

fenjalien:
This is done by checking if the syntax is correct in the `rotate` function then passing it off to `resolve-coordinate`.


johannes-wolf (approved):
:+1:

PR #31, opened 2023-05-04
Allow `rel:` in `node:` coordinates

PgBiel:
First off, thanks for the awesome library! This is absolutely huge for Typst, and fills a huge gap.

In this PR, I made it so you can specify `(node: ..., at: ..., rel: (x, y))`, in order to succinctly specify a position relative to a node's anchor. (Right now, I have to place an empty content on the node to change the "last location" pointer to do that, which isn't very pretty, haha.) I tested this locally on my project, but you may wish to perform your own tests as well.
Feel free to change the implementation or syntax.

It should be good enough for now, **but** I think we shouldn't stop here - ideally, we should make `rel:` work with other kinds of coordinates too, or even have multiple of it. But we could discuss that better in an issue before making actual progress here.

### DIFF
--- a/manual.typ
+++ b/manual.typ
@@ -70,6 +70,9 @@ There are four different ways to specify coordinates.
   + Anchor: `(node: "name", at: "example")` or `"name.example"` \
     "The position of anchor `"example"` on node with name `"name"`." \
     See @anchors
+  + Relative to anchor: `(node: "name", at "example", rel: (x,y))` \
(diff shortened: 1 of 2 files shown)


PgBiel:
Actually, I think a more general `rel` improvement is overall desired. I'll close this PR for now, as I think this would eventually become some one-off syntax. We should probably do some bikeshedding first on how things should look like.

Issue #46, opened 2023-05-22
The arc function is wrong


lecarrera-tec:
The coordinates x and y are reversed. x should use cos (instead of sin) and y should use sin (instead of cos). The files cmd.typ and draw.typ should be updated.

PR #48, opened 2023-05-25
Fix an error with the coordinates of the arc.

lecarrera-tec:
Is the issue #46 fixed. The x position should use cos while the y position should use sin.


johannes-wolf:
Some of the tests fail because the origin of the arc changed to its end.

lecarrera-tec:
I could fix local-anchor and primitive examples. I'm trying to understand the ring example, but I believe the right solution is to replace the ref.png image. With that, all the examples run correctly, well except for the image test that give me this error:
[TEST] image ...
Syntax Warning: Read from memory error. Got 0 bytes, block should be of 128 bytes
Syntax Warning: read ICCBased color space profile error

johannes-wolf:
The warning is “normal”, I also get it.
Can you push your fixes? Updating the ref image is fine.

johannes-wolf:
@fenjalien Are you using the `arc` function in `circuitypst`; is this change compatible?

fenjalien:
Thanks for the bug fix! It is technically my fault...

Go ahead with the merge, I'll make a fix to circuitypst.

PR #51, opened 2023-05-28
Use a single transform matrix

fenjalien:
This also adds matrix multiplication and inversion


johannes-wolf:
I will merge this as soon I have done some tests :)

PR #54, opened 2023-05-30
Add padding around content blocks

samuzora:
This PR adds the option of having padding around content() blocks, as per https://github.com/johannes-wolf/typst-canvas/issues/53#issue-1731344063.

Set via `content-padding(length)` or `content(..., padding: length)`.

### DIFF
--- a/draw.typ
+++ b/draw.typ
@@ -265,7 +265,7 @@
   ct,
   angle: 0deg,
   anchor: none,
-  padding: none,
+  padding: auto,
   name: none
   ) = {
   let t = coordinate.resolve-system(pt)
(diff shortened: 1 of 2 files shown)


draw.typ
```diff
@@ -256,7 +265,7 @@
   ct,
   angle: 0deg,
   anchor: none,
-  padding: 0.5em,
```
johannes-wolf: This should be `auto`. And if `auto` then use ctx-padding, otherwise use parameter.

samuzora: Okay, the latest commit sets it to auto. Thanks :)

johannes-wolf (approved):
LGTM :rocket:

PR #56, opened 2023-06-03
The info for the arrows are encapsulated inside the context.

lecarrera-tec:
The idea, is to set a separated context for the arrows, without having to specify in the lines. This would make possible to add arrows to other type of figures. In this pull, only add the option. In the short time, I will add to bezier curves during this week.

If accepted, I will update the documentation accordingly.


fenjalien:
Can you provide an example (code and image of result) of what you want to acheive? I'm not sure I fully understand...

However, as a guess, it looks like you are adding a styling method for arrows. Something similar to `stroke` and `fill`? If this is the case I would suggest a more generalised styling option like [`\tikzset`](https://tikz.dev/tikz-scopes#pgf.back/tikzset). I was going to give some ideas here but I think it would be better to open a new issue to discuss it.

lecarrera-tec:
The idea is to define the style for the arrows independently from the context, so if you want a fill for the arrow, for example in a bezier curve, that has also the option to be filled, you could have both.

In context, I am defining the "arrow" map, that has "begin", "end", "size", "stroke" and "fill". They are defined by default in canvas as:
arrow: (begin: none, end: none, size: 0.15, fill: black, stroke: black + 1pt)
Take the line as an example. I remove all the marks as arguments. The line now looks for "ctx.arrow.begin" or "ctx.arrow.end", and if some of they are set (just as you look for mark-begin or mark-end to be set), use ctx.arrow as the context for drawing the arrow.
```
#align(center, canvas({
  import "draw.typ": *
  stroke(blue + 2pt)
  line((0,0), (0,2))
  arrow(begin: ">")
  line((1,0), (1,2))
  arrow(size: 0.35)
  line((2,0), (2,2))
  arrow(end: ">")
  line((3,0), (3,2))
  arrow(begin: none, fill: blue)
  line((4,0), (4,2))
}))
```
example

PR #61, opened 2023-06-05
The arc command is updated.

lecarrera-tec:
I put aside the other pull request, to have this one ready.

The option `delta` is added, and `start` and `stop` are not obligatory. But two of the three should be defined, if not an error will be given. Fixed an error that the start angle could not be less than the stop angle (when computing the samples, the number was negative). The manual is updated and the `mode` option is added to the manual.


draw.typ
```diff
@@ -231,8 +231,11 @@
   ),)
 }
 
-#let arc(position, start, stop, radius: 1, mode: "OPEN", name: none, anchor: none, fill: auto, stroke: auto) = {
+#let arc(position, start: auto, stop: auto, delta: auto, radius: 1, mode: "OPEN", name: none, anchor: none, fill: auto, stroke: auto) = {
+  assert( ((start != auto and (stop != auto or delta != auto)) or (stop != auto and delta != auto)) and (start == auto or stop == auto or delta == auto), message: "Exactly two of three options start, auto and delta should be defined")
```
fenjalien, line 235: I think a less confusing way to check would be to use filter to get all the `auto`s and make sure there is only one. Also change "auto" to "stop" in the message

manual.typ
```diff
@@ -230,23 +230,25 @@ Draws a rectangle to the canvas.
...
-#arc(position, start, stop, radius: 1, name: none, anchor: none, fill: auto, stroke: auto)
+#arc(position, start: auto, stop: auto, delta: auto, radius: 1, name: none, anchor: none, fill: auto, stroke: auto, mode: "OPEN")
 ```
 #def-arg("position", `<coordinate>`, [The coordinate to start drawing the arc from.])
 #def-arg("start", `<angle>`, [The angle to start the arc.])
 #def-arg("stop", `<angle>`, [The angle to stop the arc.])
-#def-arg("radius", `<number>`, [The angle to stop the arc.])
+#def-arg("delta", `<angle>`, [The angle that is added to start or removed from stop.])
+#def-arg("radius", `<number>`, [The radius of the arc.])
+#def-arg("mode", `<string>`, [The options are "OPEN" (the default, just the arc), "CLOSE" (a circular segment) and "PIE" (a circular sector). The filling will depend whether the local `fill` variable is set to other that `none`, or if set to `auto`, the global context variable `fill` is set to other than `none`.])
```
fenjalien, line 242: This filling description doesn't make sense.

manual.typ
```diff
@@ -230,23 +230,25 @@ Draws a rectangle to the canvas.
...
 
 #example({
```
fenjalien, line 244: As a nice to have could you add extra examples? e.g. an arc with "CLOSE" using start and delta, and an arc with "PIE" using stop and delta.

Don't worry if you don't want to :)

fenjalien (changes requested):
Nice job! I think all of those bugs are my doing...

manual.typ
lecarrera-tec: No problem. Anyway, I have to make a little small change in the code, so PIE is closed correctly.

lecarrera-tec:
I don´t like that if you put stop and delta, the point is for the start point. But thinking about it, if you want to go backwards, then use start and -delta. I think it does not contradict the element definition from the manual.

johannes-wolf (approved):
LGTM.

fenjalien (approved):
:)

PR #71, opened 2023-06-27
Node yield

### LINKED ISSUE #70: Feature request, yielding many nodes from content
The idea would be to do something like this:
```lisp
#import "../../canvas.typ": *

#box(stroke: 2pt + red, canvas({
    import "../../draw.typ": * 
    content([ #content_node("lab1")[This] is #content_node("lab2")[Sparta]  ]),
    line("lab1", "lab2") 
}))
```

where there would be a line drawn from `This` to `Sparta`
tikz has support for something like this, and it would help a lot cause doing this by hand is very annoying/not reliable

I'm looking at the code trying to see how feasible it would be.

with maybe `cn` or `node` or `cnode` as an alias or replacement for content_node

astrale-sharp:
Closes #70


astrale-sharp:
# What's the plan
create a new node function in draw that can be called from inside a content block storing info of name and bounds and anything that's needed to create anchors.

It gets coerced as content and is in the children of our initial content block.

We parse it/eval it from it's representation, get back the info we need during the draw.content call and use it to modify the ctx with new anchors

# How far I am in the implementation
The test looks like this: 
```lisp
#import "../../canvas.typ": *

#{
    import "../../draw.typ": * 
    let x = content(())[This #node()[is] a #node(name : "brown")[*_brown_*] fox]
    
    [#hide_nodes(x.at(0).ct) \ \ ]

    [#get_nodes(x.at(0).ct)]
}
```

image

# What's blocking me right now

To call measure in node() you need an access to style which you can't do from inside the content block with without converting your dictionary to content, it might not be blocking since this already happens, but the parse_eval implementation need to be able to parse the result of measure. (this will be my next try)

It is probably needed although it could be called during the draw.content call when ctx can be accessed which will be what I'll try if the precedent approach fails.

Issue #72, opened 2023-06-28
Start using semver / Submit package


fenjalien:
Given Typst has started taking package submissions we should obviously submit this one!

- [x] Add a `typst.toml`
- [x] Sort out versioning
  - I would advise we start at `0.0.1` and only increase maintanence number (e.g. `0.0.2`) until we are happy that this package is "stable" and won't make any more drastic changes.
- [ ] Create a github actions release workflow
  - This should only keep what is related to the running of the package (e.g. no manual or tests)

I'm not sure *when* we should release a new version, easiest would be on every PR merge...

johannes-wolf:
There is the question how the `#import "typst-canvas/draw.typ": *` can be realized with the package manager.

One option would be having an additional typst file that imports `canvas.typ` and `draw.typ`.
Downside is, that one has to use all commands with a namespace.

fenjalien:
Hmm could point, that might be one for the questions channel in the discord server...

fenjalien:
Okay it shouldn't be a problem if we do it like this:
`lib.typ`:
```
#import "canvas.typ": canvas
#import "draw.typ": *
#import "vector.typ"
#import "matrix.typ"
```
`main.typ`:
```
#import "@local/typst-canvas:0.0.1": canvas

#canvas({
  import "@local/typst-canvas:0.0.1": *
})
```

johannes-wolf:
The PR #73 uses the following approach:
```
#import "@local/canvas:0.0.1"

#canvas.canvas({
  import canvas.draw: *
  // ...
})
```

N3xed:
Just for some inspiration, I use it like this:

```
#canvas({
    import canvas-draw: *
})
```

so `lib.typ` would be e.g.:
```
#import "canvas.typ": canvas
#let canvas-draw = (() => {
    import "draw.typ"
    return draw
})()
```

But maybe they'll add support for arbitrary properties on functions for non-builtin functions, then `draw` could be a property on the `canvas` function (which would be nice).

johannes-wolf:
@N3xed I would like to stick to the `import canvas.draw` version. You can then write:
```
#import '@local/canvas:0.0.1': canvas, draw
#canvas({
  import draw: *
  // ...
})
```

N3xed:
My only concern with that is, that `draw` is a really common function name. And with your approach `draw` would be a globally available module that is only used to import the actual draw functions. I guess what I'm trying to say is, that I think it doesn't really fit the name.

[...]

PR #73, opened 2023-06-30
Support typst package

johannes-wolf:
Initial support for typst package infrastructure.

This PR adds a `package` script, that allows building a package of `typst-canvas` and also installing it locally.

Partly implements #72.


README.md
```diff
@@ -1,8 +1,44 @@
...
+For information, see the [manual](manual.pdf).
+
+### Just
+
+This project uses [just](https://github.com/casey/just), a handy command runner.
+
+You can run all commands without having `just` installed, just have a look into the `justfile`.
+To install `just` on your system, use your systems package manager. On Windows, [Chocolatey](https://chocolatey.org/) (`choco install just`) can be used.
+
+## Installing
```
fenjalien, line 16: If this is the exact readme that we want to submit there should be something about using the package manager. Espcially as this would be the way the majority of users will be importing the package.

README.md
```diff
@@ -1,8 +1,44 @@
...
+## Testing
+
+This package comes with some unit tests under the `tests` directory.
+To run all tests you can run the `just test` target.
+
+You need to have ImageMagick installed on your system, which is needed for image comparison.
+
+### Windows
+If you are using the [Chocolatey](https://chocolatey.org/) package manager, you can install imagemagick using `choco install imagemagick`.
+Otherwise download and install a matching package from the [ImageMagick](https://imagemagick.org/script/download.php) website.
```
fenjalien, line 44: Maybe just include a link to their main site like you've done with `just` and give a short description of what its used for.

scripts/package
fenjalien, line 1: It should be optional to have the script place it in the correct folder for the user. We would also want to use this in a github workflow for releasing

scripts/package
```diff
@@ -0,0 +1,49 @@
...
+  collisions.typ
+  coordinate.typ
+  LICENSE
+)
+
+function read-toml() {
+  local file="$1"
+  local key="$2"
+  awk -F= "/^${key}/ { sub(/^\\s*\"/, \"\", \$2); sub(/\"\\s*$/, \"\", \$2); print(\$2); }" "$file"
+}
```
fenjalien, line 30: What does this do?

typst.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "canvas"
+version = "0.0.1"
+repository = "https://github.com/johannes-wolf/typst-canvas"
+entrypoint = "lib.typ"
+authors = [
+  "Johannes Wolf <https://github.com/johannes-wolf>",
+  "fenjalien <https://github.com/fenjalien>"
+]
+license = "Apache-2.0"
+description = "A library for drawing with typst with an API inspired by Tikz and processing."
```
fenjalien, line 11: Make sure to capitalise **T**ypst and Processing. 

This description should ideally be the same as the one at the top of the readme and in the description of the repo. I do like this one better though.

fenjalien (changes requested):
Woop can't wait for this. Are we having a github workflow for release though?

README.md
johannes-wolf: For ImageMagick? The reason its needed is in the prev. paragraph.

README.md
```diff
@@ -1,8 +1,44 @@
-# Typst-Canvas
-A simple [Typst](https://typst.app) library for drawing stuff.
+# Canvas
 
-For more information, see the [manual](manual.pdf).
+Canvas is a simple [Typst](https://typst.app) library for drawing stuff.
 
-## Demos
-### Chart - [PDF](demo/3d-diagram.pdf) - [TYP](demo/3d-diagram.typ)
-[![Bar Chart](demo/3d-diagram.pdf.png)](demo/3d-diagram.typ)
```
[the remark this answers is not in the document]
johannes-wolf: Yes. But we need something nicer :) than the one I have removed. The Karl picture might be a bit too big? Maybe we create a smaller version or put multiple scaled down images next to each other?

johannes-wolf:
> Woop can't wait for this. Are we having a github workflow for release though?

No workflow yet. Should the workflow output a zip for submission?

scripts/package
johannes-wolf: It is optional. You can pass a custom path to place the output at.

scripts/package
johannes-wolf: Parse the version from the toml.

scripts/package
fenjalien: I mean by default it should create a folder in the current directory. Maybe call the script and command "bulid" and have a `--install` flag which places it in `@local`. Can you also document that behaviour please.

scripts/package
fenjalien: Can you add a comment saying that

README.md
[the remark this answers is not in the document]
fenjalien: I might have a few ideas for this actually

scripts/package
johannes-wolf: I have added a help to the script that explains what its doing.
Either invoke it without arguments or with `help`.

The justfile wraps the invocation by having the `install` target and the `package <path>` target.

README.md
```diff
@@ -1,8 +1,59 @@
...
+To install `just` on your system, use your systems package manager. On Windows, [Chocolatey](https://chocolatey.org/) (`choco install just`) can be used.
+
+## Testing
+
+This package comes with some unit tests under the `tests` directory.
+To run all tests you can run the `just test` target.
+
+You need to have ImageMagick installed on your system, which is needed for image comparison.
+
+### Windows
```
fenjalien, line 54: Okay so to get `just` working on my windows machine I installed it via cargo then used it in git bash. In their book there appears to be multiple ways to install it for windows and pre-built binaries so maybe link to those pages. Then recommend running it in git bash as it comes optional with git for windows?

just manual: https://just.systems/man/en/

johannes-wolf: I have added the Cargo option + a link to the `just` documentation about the packages.

fenjalien:
The workflow should just output a folder right?

johannes-wolf:
> The workflow should just output a folder right?

Tried a release workflow but it gives no benefit, as releasing to typsts packages repo is a manual step.
If we want a release workflow, I suggest adding that later, as I think it is less important. Also doing things with github-actions is painful imho.

README.md
```diff
@@ -1,8 +1,84 @@
...
+      <img src="gallery/pie-chart.png" width="250px">
+    </a>
+  </td>
+  <td>
+    <a href="gallery/3d-chart.typ">
+      <img src="gallery/3d-chart.png" width="250px">
+    </a>
+  </td>
+</tr><tr>
+  <td>Karls Picture</td>
```
fenjalien, line 24: Karl's Picture

fenjalien (changes requested):
Just remembered about the examples in the manual, they should be updated!

PR #75, opened 2023-06-30
Improve test script

### LINKED ISSUE #74: Improve Test Script
The current test script only works on Linux and I use Windows so I have to make some changes when testing.

Objectives for a new test script (and future scripts in general):
- [ ] Cross platform
- [ ] Easy dependencies for the user to install
  - Provide clear instructions and links to exact sites / downloads
- [ ] Don't affect the file system outside of the repository's folder
  - I *really* don't like that the test script wants to install Typst again while I already have it obvsly
Optional:
- [ ] Re-run failed tests

Maybe we could try using python?

johannes-wolf:
Changes
- Removes dependency on pdftoppm and uses typsts native png export
- Supports image magick with and without the `magick` binary
- Tests do not auto install typst binary (pass `--install` to install typst under `/tmp`, linux only!)

Partly fixes #74.


fenjalien (approved):
Okay took me a bit but this seems to work without any changes for me! I ended up installing `just` through `cargo install` and ran it in git bash (I don't remember if I've modified it).

PR #85, opened 2023-07-10
Improve Content Rotation

johannes-wolf:
This PR improves content rotation:
- Allow `angle` argument to be of type `coordinate` to use an angle between two coordinates
- Calculate rotated anchors to allow positioning labels on/above/below paths/lines

It also adds:
- Documentation for `merge-path`
- A new utility function `for-each-anchor` that allows iterating a nodes anchors


johannes-wolf:
@fenjalien What do you think about the `angle` option?

PR #92, opened 2023-07-21
Add Plotting + Chart Library

johannes-wolf:
This PR adds a plotting `plot`, `chart`, `axes` and `palette` library.


manual.typ
```diff
@@ -1220,3 +1245,165 @@ for-each-anchor("my-rect", (name) => {
   }
 })
 ```)
+
+= Libraries
+== Plot
+
+The plotting library `plot` of CeTZ allows drawing line charts of data points.
+All data is expected to be in `(x, y)` format.
```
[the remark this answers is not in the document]
johannes-wolf: Specified in the new version.

manual.typ
```diff
@@ -1209,7 +1234,7 @@ The example below shows how to use this system to create an offset from an ancho
             angle: -45deg)
     }
   })
-}, ```typ
+}, ```typc
```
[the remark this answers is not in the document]
johannes-wolf: Will do.

manual.typ
```diff
@@ -1220,3 +1245,165 @@ for-each-anchor("my-rect", (name) => {
...
+#def-arg("data", "a|function",
+  [The data point array or callback function])
+#def-arg("style", "style",
+  [Style dictionary used for drawing (same as for `set-style`)])
+#def-arg("epigraph", "b",
+  [If true, fill the epigraph of the function (style key: `epigraph`)])
+#def-arg("hypograph", "b",
+  [If true, fill the hypograph of the function (style key: `hypograph`)])
+#def-arg("fill", "b",
+  [If true, fill the function (style key: `fill`)])
```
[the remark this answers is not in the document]
johannes-wolf: I was wondering how to use `style` in code where we have no context, as we can not get inherited style values. This means using style would not work as expected.

I have changed this a bit in the new version.

manual.typ
```diff
@@ -1220,3 +1245,165 @@ for-each-anchor("my-rect", (name) => {
...
+== Plot
+
+The plotting library `plot` of CeTZ allows drawing line charts of data points.
+All data is expected to be in `(x, y)` format.
+
+=== Data <plot-data>
+
+Data can be passed as array of data points or dictionary to provide further information. The data dictionary allows for the following keys:
+#def-arg("data", "a|function",
+  [The data point array or callback function])
```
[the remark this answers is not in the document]
johannes-wolf: For the new API I tried to describe it by writing "... a function in the form `x => y`, I hope this is more clear?

manual.typ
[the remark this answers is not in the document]
fenjalien: I was thinking a bit more about this and I think we have a `children-ctx` key that should be a function that accepts ctx. We could allow a group's body be that function.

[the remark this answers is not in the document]
johannes-wolf: This is already possible, by just passing a function.
But how should e.g. epigraph be set? like `set-style(plot: (epigraph: red))`?

manual.typ
```diff
@@ -593,10 +588,13 @@ merge-path({
...
 == Groups <groups>
 Groups allow scoping context changes such as setting stroke-style, fill and transformations.
-```typ
-#group(content, name: none)
+
+```typc
+group(content, name: none, anchor: none)
 ```
 
+Note: You can pass `content` a function of the form `ctx => array` that returns the groups children. This way you get access to the groups context object.
```
fenjalien, line 596: I don't think having the return type `array` is very clear as we don't explain anywhere that actually the draw functions return arrays that implicitly get joined. I also think we should rename `content` to `body` as you can't use Typst's content. Maybe just say "`body` can also be function takes a single parameter, a dictionary that describes the group's current context such as styling".

We should also start making some separate developer documentation for the context dictionary but I'm not to bothered about it right now.

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+```typc
+plot(size: (width, height),
+     axis-style: "scientific"
+     ..options,
+     body)
+```
+
+The ticks option dictionary supports the following keys:
+#def-arg("size", "a",
+  [Size of the plot])
```
fenjalien, line 1397: Give more specification of the expected array. Types, length, what each position means etc.

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+  / label : The axis label
+  / min: Axis min. value
+  / max: Axis max. value
+  / ticks: List of tick values or value/label tuples
+  / tick-step: Tick step or `none`
+  / minor-tick-step: Minor tick step or `none`
+  / decimals: Number of tick label decimal digits
+  / unit: Tick label suffix
+  ]
+])
```
fenjalien, line 1412: Could the axes use a similar input to styling? As in nested dictionaries:
```
plot(
  axes: (
    x: (
      label: "X",
      ...
     ),
    y: ...
  )
)
```
You could then allow options to be applied to both axes by allowing them in the base of the `axes` dict or through a `default` dict.
```
plot(
  axes: (
   // Applies to all axes
  ticks: ...
   // or if you want to keep axis names clear
   default: (
      ticks: ...
    ),
   // An axis
    y: ...
  )
)
```
This might even allow them to be set through `set-style`?t

fenjalien: You might also want to group tick options like step, minor-step, values etc.

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+#def-arg("body", "..",
+  [Calls of `plot.add(..)`, see <plot-add>])
+
+=== Plot-Add
+
+The `plot.add` function adds plotting data into a plot environment.
+It must be called from insides `plot({..})`.
+
+If used with a callback, the `domain` must be specified!
```
fenjalien, line 1421: callback? what callback?

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+
+```typc
+add(domain: auto, hypograph: false, epigraph: false, fill: false,
+    mark: none, mark-size: .2, mark-style: (:), samples: 100,
+    style: (:), axes: ("x", "y"),
+    data)
+```
+
+#def-arg("domain", "a|auto",
+  [Range of x for sampled data])
```
fenjalien, line 1431: What happens when domain is auto? Explain the expected array types, length and each position means

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+#def-arg("mark", "s?",
+  [Mark symbol. Any of `("x", "o", "square", "triangle", "|", "-")`])
+#def-arg("mark-size", "f?",
+  [Size of mark symbol])
+#def-arg("mark-style", "style?",
+  [Style used for drawing marks. Note that this inherits the plots style])
+#def-arg("style", "style",
+  [Style used for drawing the graph])
+#def-arg("samples", "i",
+  [Number of samples to use (only if `data` is of type function)])
```
fenjalien, line 1447: Could any of these be absorbed into styles?

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+#def-arg("axes", "a",
+  [Array of axis names to use])
```
fenjalien, line 1449: Use for what?

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+#def-arg("data", "function|a",
+  [Array of data points or function in the form `x => y`])
```
fenjalien, line 1451: What format is expected array supposed to take? length, types, position etc. What are the expected types of `x` and `y`?

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+          x-unit: $degree$, y-max: .5, {
+  plot.add(domain: (0, 360), x => calc.sin(x * 1deg))
+  plot.add(domain: (0, 360), x => calc.cos(x * 1deg),
+           samples: 10, mark: "x", mark-style: (stroke: blue))
+})
+```)
+
+== Chart
+
+With the `chart` library it is easy to draw charts.
```
fenjalien, line 1485: I understand that they are fundamentally different but would it make sense for `chart` and `plot` to be merged in some way? This isn't a suggestion just wondering.

fenjalien, line 1485: Give a short list here of the types of charts currently supported

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+barchart(size: (width, height))
+```
+
+#def-arg("data", "a",
+  [Data array of arrays or dictionaries])
+#def-arg("label-key", "s|i",
+  [Data item key to access an items label])
+#def-arg("value-key", "s|i|a",
+  [Data item key(s) to access an items value(s). For multi-value charts
+   this must be an array of all keys, e.G. `(..range(1, 5))`])
```
fenjalien, line 1499: I think the format of the data needs a bit more information as the label and value keys don't make sense.

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+#def-arg("mode", "s",
+  [Barchart mode, one of "basic", "clustered" (bars next to each other), "stacked" (bar stacked) or "stacked100" (bars stacked but as percentage of their sum)])
+#def-arg("size", "a",
+  [The chart's size. Height can be set to `auto`.])
```
fenjalien, line 1503: array types, length, position meanings

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+#def-arg("x-tick-step", "f",
+  [X axis tick step])
+#def-arg("x-ticks", "a",
+  [X axis ticks list, a list of tick values or value/label tuples])
+#def-arg("x-unit", "content",
+  [X axis tick label suffix])
+#def-arg("x-label", "content?",
+  [X axis label])
+#def-arg("y-label", "content?",
+  [Y axis label])
```
fenjalien, line 1518: Could these be absorbed into a `styles`?

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+==== Stacked
+#example({
+  let data = (("A", 10, 12, 22), ("B", 20, 1, 7), ("C", 13, 8, 9))
+  chart.barchart(size: (3, auto), mode: "stacked",
+                 x-tick-step: 10, value-key: (..range(1, 4)), data)
+}, ```typc
+let data = (("A", 10, 12, 22), ("B", 20, 1, 7), ("C", 13, 8, 9))
+chart.barchart(size: (3, auto), mode: "clustered",
+               x-tick-step: 10, value-key: (..range(1, 4)), data)
+```)
```
fenjalien, line 1550: If you want some more room for these examples, `examples` has a vertical option which places the drawing above the code

manual.typ
```diff
@@ -1415,3 +1374,187 @@ tree.tree(data, content: (padding: .1), direction: "right",
...
+
+== Palette <palette>
+
+A palette is a function that returns a style for an index.
```
fenjalien, line 1554: This is a pretty cool addition, an explantion of how to create one with an example wouldn't go amiss

tests/chart/test.typ
```diff
@@ -0,0 +1,62 @@
...
+).rev()
+
+#let data2 = (
+  ([15-24], 18.0, 20.1, 23.0, 17.0),
+  ([25-29], 16.3, 17.6, 19.4, 15.3),
+  ([30-34], 14.0, 15.3, 13.9, 18.7),
+  ([35-44], 35.5, 26.5, 29.4, 25.8),
+  ([45-54], 25.0, 20.6, 22.4, 22.0),
+  ([55+],   19.9, 18.2, 19.2, 16.4),
+).rev()
```
fenjalien, line 21: why is the data being reversed?

tests/plot/ref.png
fenjalien, line 1: These are some really funky graphs, great work :)

gallery/barchart.typ
```diff
@@ -0,0 +1,22 @@
...
+#set page(width: auto, height: auto, margin: .5cm)
+
+#let data2 = (
+  ([15-24], 18.0, 20.1, 23.0, 17.0),
+  ([25-29], 16.3, 17.6, 19.4, 15.3),
+  ([30-34], 14.0, 15.3, 13.9, 18.7),
+  ([35-44], 35.5, 26.5, 29.4, 25.8),
+  ([45-54], 25.0, 20.6, 22.4, 22.0),
+  ([55+],   19.9, 18.2, 19.2, 16.4),
+).rev()
```
fenjalien, line 12: why is this data being reversed?

fenjalien (changes requested):
The api feels a lot cleaner now and wow thoughs graphs :) However it would be good to get the styling to work in a similar manner to `set-style`.

I've focused my review mostly on the design of the api in the manual. I don't mind how its been implemented in the lib files unless there is something specific you want me to look over.

tests/plot/ref.png
johannes-wolf: :blush:

manual.typ
johannes-wolf: They are both different front-end APIs to `axes` + some drawing.
An area or linechart may use `plot` internally. But I would like to keep them separate.
Plot's intention is (2D) function plotting, while chart's is for creating out-of-the-box good looking diagrams with most things handled by the lib.

manual.typ
johannes-wolf: Yeah, but we should first generate function docs from the code doc. I do not like having both.

manual.typ
johannes-wolf: I am not sure if that is good, as it comes with no real benefit.

Issue #143, opened 2023-08-26
styling: Fix `auto` inheriting "old" attributes


johannes-wolf:
The following example draws a red line with an red arrow (mark):
```
set-style(stroke: red)
line((0,0), (1,0), mark: (end: ">"))
```

This example on the other hand, draws a blue line with a (still) red arrow head:
```
set-style(stroke: red)
line((0,0), (1,0), mark: (end: ">"), stroke: blue)
```

This is, because the `auto` stroke attribute of `mark` gets set to the root `stroke` of the context.
But I expect it to get set to the attribute set as style (blue). @fenjalien Any thoughts on this?

fenjalien:
In a way this is expected behavior as that mark style inherits from the root style.
It could be changed so that styles inherit directly but that would cause a similar issue in the other direction. you would need to specify the stroke of mark if you want it different to the stroke of the line.

I think the best way would be to allow a style value that causes direct inheritance instead of from the root.
- `auto` inherits from the root style
- `"inherit"` inherits from the styles direct parent.

johannes-wolf:
In which case do you expect some style not to inherit the "root" style instead of the one just set, though?

fenjalien:
The style should inherit as much as it could from its parent, if it does not exist on the parent it should inherit from the style's root.

e.g.
```
(line: (mark: "inherit", ...))
```
`mark` should inherit as much as it can (stroke, fill) but the rest of the mark styling for line should take on the root style of mark (arrows...).
```
(line: (mark: (stroke: "inherit")))
```
the mark's stroke should inherit from line but all other mark styling should be the root mark style (fill..)

I think the best way to look at it is which direction we traverse the style dictionary to find the first instance of what to inherit from. `auto` is into the dictionary from the root, `"inherit"` is up from the current position.

johannes-wolf:
With the current implementation `auto` only inherits form the very "bottom" though.
But the differentiation between `auto` and `inherit` makes no sense to me.
When do you want this:
```
set-style(stroke: red)
line(.., stroke: blue, mark: (end: ">"))
```
have red marks? Therefore I propose to always inherit top-down instead of bottom-up, like in CSS.

johannes-wolf:
Link #165

johannes-wolf:
Fixed with #200.

PR #145, opened 2023-08-27
style: Add "inherit" style attribute to inherit the current style

johannes-wolf:
#143 @fenjalien Please have a look at this. This is a fix for #143 and #147. I would like to do a release after this merge.


src/styles.typ
fenjalien, line 1: I'm not entirely sure how this works, so some comments might be needed. Even so I think this implementation could be better, but if its just to get it to work while the styling is re-worked I'm happy to let it slide...

src/styles.typ
```diff
@@ -23,21 +57,33 @@
   if root != none {
```
fenjalien, line 57: Is this loop necssary if `inherit` exists? Or could the logic in `inherit` be put here instead?

src/styles.typ
```diff
@@ -23,21 +57,33 @@
...
   }
   return current
 }
 
+/// Resolve style
+///
+/// - current (style): Current style dictionary
+/// - new (style): Style to merge onto current
+/// - root (string,none): Style root key
+/// - inject (style, none): Inject new style at `root`
```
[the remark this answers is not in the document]
johannes-wolf: It is for libraries for injecting their custom style defaults. Maybe we could only do that once if the key does not exist at all in the style. I will change that.

johannes-wolf:
Should we just close this and come up with a new styling system? Maybe a simpler one? Fixing the current one is extremely frustrating, I am running out of ideas, it always breaks. Maybe we do not need a nested system but can use flat keys, like CSS does? This would simplify things a lot.

fenjalien:
I'm working on a proposal for the new system...

PR #164, opened 2023-09-09
Improve mark positioning

### LINKED ISSUE #162: draw: Improve arrows
- Try calculating the miter (depending on miter style and width) 
- Add options for out/insetting arrow heads

Example of Tikz arrows:
grafik

Example in CeTZ:
grafik

johannes-wolf:
Idea: Calculate the miter of joined strokes and inset triangle marks along their direction.

*This is only implemented for marks drawn using line(..)*, because if we offset the mark, we also have to shorten the line. Implementing this for beziers and `place-marks` is possible, but not trivial. What do you think @fenjalien @Andrew15-5?

Fixes #162


CHANGES.md
```diff
@@ -13,6 +13,9 @@
 - Added `mark: (..)` support for `bezier` and `bezier-through`
 - Changed `content` to allow for a second coordinate to span content between
 - Added `content: (frame: "rect"|"circle")` style for drawing a frame around content
+- Default triangle mark angle changed. It can now be set via the style attribute `angle`.
+- Arrowheads on lines are now offset so the tip of the triangle points exactly on the
```
Andrew15-5, line 17: There is a trailing whitespace.

Andrew15-5, line 17:
```suggestion
- Arrowheads on lines are now offset so the tip of the triangle points exactly on the
```

Andrew15-5:
Please, show me a snippet, where I can see the difference. I've used my library with no config except `fill`, `end` and `thickness`. It still shows the same:

image

I'm using `@local` namespace with `mark-improvements` branch.

johannes-wolf:
@Andrew15-5 Something like `line((0,0), (1,0), mark: (end: ">")); rect((1,-1), (2,1))` should be enough to test it.

johannes-wolf:
@fenjalien I think we should redo arrows. They should behave the same for lines and beziers (and custom paths), which means we have to implement a way to shorten beziers (which is not that hard).

Andrew15-5:
I think I was trying to do this through my local package, but I did a pretty bad decision of putting the same import line for the `cetz` package in every single file of the flowchart symbol. I now only import it once in the `utils.typ` and then import it through that file. And the relevant issue was that in every file the import was from `preview`, but I should've changed it to the `local`. Now I fixed both and it looks pretty sweet:

image

johannes-wolf:
@Andrew15-5 So this works for you? I'd like to improve the behaviour for other kinds of paths (combined, bezier) too, so i won't merge this yet.

PR #181, opened 2023-09-13
layer: Introduce basic layer/z-index support

### LINKED ISSUE #178: draw: Support layers
Support drawing behind already drawn nodes. This could be in the form of an `after(top, bottom)` function, that first draws `top`, then `bottom` but orders all drawing commands from bottom before the drawing commands of `top`.

Motivation: being able to draw behind elements while knowing their size/anchors.

johannes-wolf:
This allows setting a z-index on elements by wrapping them in `on-layer(index, body)`.

Fixes #178.


manual.typ
```diff
@@ -441,6 +440,25 @@ for-each-anchor("demo", (name) => {
 })
 ```
 
+== Layers
+
+You can use layers to draw elements below or on top of other elements by using layers
+with a z-index. When rendering, all draw commands are sorted by their z-index (0 being
+the default).
```
fenjalien, line 447: Can we stick with the `layers` name? `z-index` could get confusing when we also allow a `z` component.

src/canvas.typ
```diff
@@ -316,13 +316,17 @@
...
     }
-
   }
 
   if bounds == none {
     return []
   }
 
+  // Order draw commands by z-index
+  draw-cmds = draw-cmds.sorted(key: (cmd) => {
```
fenjalien, line 326: This is a really simple way to do this, nice. But is it the best way given all drawables are looped even if the majority of them will be on the 0 layer?

src/draw.typ
```diff
@@ -306,6 +306,30 @@
   ),)
 }
 
+/// Draw body on z-index layer
+///
+/// This can be used to draw elements behind already drawn elements by
+/// using a lower layer value (i.e -1).
+/// 
+/// - layer (integer, float): Z-Index to draw the children at
```
fenjalien, line 314: why a float?

src/draw.typ
```diff
@@ -306,6 +306,30 @@
...
+/// - body (draw, function): Child elements or function (`ctx => draw-cmds`)
```
fenjalien, line 315: Do we have a dedicated type description to mean "elements", if not could you add one and use it in `canvas` and where ever else it needs.

src/draw.typ
```diff
@@ -306,6 +306,30 @@
...
+/// - force (bool): Foce child elements on a different layer to this layer
```
fenjalien, line 316: I do not understand what this means

fenjalien (changes requested):

src/draw.typ
johannes-wolf: Why not?

src/canvas.typ
johannes-wolf: I see no problem with that.

src/draw.typ
johannes-wolf: I removed that option. I was about moving from a layer to a different one, as the `layer` function does not overwrite an elements layer to support nesting.

src/draw.typ
johannes-wolf: We use elements at other places, I changed it to that.

src/draw.typ
fenjalien: To me layers should be discrete not continuous but I guess it doesn't really effect much.

manual.typ
fenjalien: I don't see these changes?

src/canvas.typ
fenjalien: If you say so :)

manual.typ
johannes-wolf: Oops, I had changed the code comment but missed the docs :)

PR #189, opened 2023-09-15
style: A new style system

johannes-wolf:
**This is a Breaking Change** that break every current use of `set-style`.
Migration is hard, because the current style system does not know which dictionary key represents an element and which does not. Therefore I decided to panic if an "old" `set-style` invocation is detected.

This implements a selector based style system.
A style consists of a selector and a property dictionary.
Styles are sorted by their selector depth, so more generic styles get applied before more specific ones. Selectors only support element names, but could be changed to allow for more things.

`set-style(line: (stroke: green))` becomes `set-style(stroke: green, selector: "line")`
**TODO:** is `selector:` too long?

To change the mark style for beziers:
`set-style(stroke: green, size: .5, selector: "bezier > mark")`

There is also a new function to "nest" styled elements:
` style-element("outer", {...})` All child elements query their style using `"outer > ..."`.
**TODO:** The name `style-element` might not be good.

I would like to get some feedback on this.
Ping: @jneug what do you think?


jneug:
Looks interesting. Without looking at the code I already habe some questions:

- does this mean `set-style` can only change one style at a time? Before I could set styles for multiple style roots. 
- I'm assuming, calling `set-style` without `selector` sets a "root" style? Maybe make the selector a mandatory positional argument and add a special selector for "root" styles (like `*`)?
  ```typst
  // set bezier mark style
  set-style("bezier > mark",
    stroke: red)
  // set all strokes
  set-style("*",
    stroke: red)
  ```
- Why use the `>` syntax? Is this inspired by CSS selectors? Why not use dot-notation for nested styles?
  ```typst
  set-style(stroke: red
    selector:"bezier > mark")
  // vs
  set-style(stroke: red
    selector:"bezier.mark")
  ```

I'm currently busy with work, but I will give this a closer look, when I got some time to spare.

fenjalien:
Before any further work is put into this branch, can we discuss this design in #165? It would help to keep the conversation in one place and we won't have to waste as much work if we don't like the design. @johannes-wolf could you add a breif there about this design, and why not / comments on my proposal?

I do have a few thoughts and suggestions about this design but again would prefer to discuss it in the relevant place :)

Issue #185, opened 2023-09-14
set-style(mark: (end: ">")) fails with bezier curve


raphCode:
```
#import"@preview/cetz:0.1.1"

#cetz.canvas(length:1cm, {
  import cetz.draw: *
  set-style(mark: (end: ">"))
  bezier((0,0), (2, 2), (1, 0))
})
```
error: `expected color or none, found auto`

Expected result:
![image](https://github.com/johannes-wolf/typst-canvas/assets/15750438/9f11a061-af2a-4f00-9078-19bbe544739e)

johannes-wolf:
I have opened a PR to change the styling system. This would resolve this bug, too.
Please have a look here: #189

johannes-wolf:
Fixed with #200

PR #199, opened 2023-09-20
style: Fixes

### LINKED ISSUE #143: styling: Fix `auto` inheriting "old" attributes

fenjalien:
- Fixes #143 and #185 
- Manual has been updated for `"inherit"` keyword

@johannes-wolf what tests would you recommend updating/adding?


johannes-wolf:
@fenjalien The tests added with this PR: https://github.com/johannes-wolf/cetz/pull/145

PR #200, opened 2023-09-20
style: Fix the current styling system

johannes-wolf:
This should hopefully fix the current styling issues!

Note, that the default style now _must_ have all its dictionaries expanded! (stroke must not be `auto`, but the full stroke dict). We could add a function for auto-expanding the style once if needed.


fenjalien (approved):
If you're happy that it fixes the style while we overhaul, its good!

PR #209, opened 2023-09-22
aabb: Move bounding box calculation into aabb.typ

johannes-wolf:
Simplify AABB calculation and move it to its own file.
Also support 3 dim vectors (for later use in some draw commands).

### DIFF
--- a/src/aabb.typ
+++ b/src/aabb.typ
@@ -1,0 +1,57 @@
+#import "vector.typ"
+
+/// Compute an axis aligned bounding box (aabb)
+/// for a list of vectors.
+///
+/// An aabb dictionary has the following keys:
+///  - low  (vector) Min. bounds vector
+///  - high (vector) Max. bounds vector
+///
+/// - pts (array): List of vectors/points
+/// - init (dictionary): Initial aabb
+/// -> dictionary AABB object
+#let aabb(pts, init: none) = {
+  let bounds = init
+
+  if type(pts) == array {
+    for (i, pt) in pts.enumerate() {
+      if bounds == none and i == 0 {
+        bounds = (low: pt, high: pt)
+      } else {
+        let (x, y, z) = pt
+
+        let (lo-x, lo-y, lo-z) = bounds.low
+        bounds.low = (calc.min(lo-x, x), calc.min(lo-y, y), calc.min(lo-z, z))
+
+        let (hi-x, hi-y, hi-z) = bounds.high
+        bounds.high = (calc.max(hi-x, x), calc.max(hi-y, y), calc.max(hi-z, z))
+      }
+    }
+    return bounds
+  } else if type(pts) == dictionary {
+    if init == none {
+      return pts
+    } else {
+      return aabb((pts.low, pts.high,), init: bounds)
+    }
+  }
+
+  panic("Expected array of vectors or bbox dictionary, got: " + repr(pts))
+}
+
+/// Get the mid-point of an aabb as vector.
+///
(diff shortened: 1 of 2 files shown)


src/aabb.typ
```diff
@@ -0,0 +1,61 @@
...
+/// - pts (array): List of points
+/// - init (dictionary): Initial aabb
+/// -> dictionary AABB object
+#let aabb(pts, init: none) = {
+  let bounds = init
+
+  if type(pts) == array {
+    for (i, pt) in pts.enumerate() {
+      if bounds == none and i == 0 {
+        bounds = (lo: pt, hi: pt)
```
fenjalien, line 17: Not sure about `lo` and `hi`, maybe `min`/`max` or `low`/`high`?

src/aabb.typ
```diff
@@ -0,0 +1,61 @@
...
+        let (hi-x, hi-y, hi-z) = bounds.hi
+        bounds.hi = (calc.max(hi-x, x), calc.max(hi-y, y), calc.max(hi-z, z))
+      }
+    }
+    return bounds
+  } else if type(pts) == dictionary {
+    if init == none {
+      return pts
+    } else {
+      return aabb((pts.lo, pts.hi,), init: bounds)
```
fenjalien, line 36: why not unpack the dictionary at the start of the funtion?

src/aabb.typ
```diff
@@ -0,0 +1,61 @@
+/// Compute an axis aligned bounding box (aabb)
+/// for a list of points.
+///
+/// An aabb dictionary has the following keys:
+///  - lo (vector) Min. bounds vector
```
fenjalien, line 5: Do we have a formal description of what a `vector` and a `point` type is?

src/aabb.typ
```diff
@@ -0,0 +1,61 @@
...
+#let mid(bounds) = {
+  let (lo-x, lo-y, lo-z) = bounds.lo
+  let (hi-x, hi-y, hi-z) = bounds.hi
+  
+  return (lo-x + (hi-x - lo-x)/2,
+          lo-y + (hi-y - lo-y)/2,
+          lo-z + (hi-z - lo-z)/2)
+}
+
+/// Get the size of an aabb as vector
```
fenjalien, line 53: can you be more specific that its the side lengths?

fenjalien (changes requested):
its a good thing to move tbh. Although `aabb.aabb(...)` feels weird but its still a good name :)

src/aabb.typ
johannes-wolf: Not really. We use `vector` for arrays of numbers and `coordinate` for a cetz coordinate. Point should not be used as a type.

johannes-wolf:
> its a good thing to move tbh. Although `aabb.aabb(...)` feels weird but its still a good name :)

Yeah, it feels weird. Could also name it `from-points(..)`?

src/aabb.typ
fenjalien: why not add one in `vector.typ`?

johannes-wolf: > why not add one in `vector.typ`?

What do you mean?

fenjalien: Add a description of the `vector` type there, given most should be created from there?

johannes-wolf: We already use vectors everywhere but on the user API.
I actually think it is obvious that a vector is an array of scalars.

fenjalien (approved):

PR #211, opened 2023-09-23
draw: Use ellipse anchors for circle and arcs

johannes-wolf:
Use circle anchors (top/bottom-left/right at 45deg) for both ellipses and arcs (see ref images).


fenjalien (commented):
Looking at https://tikz.dev/library-shapes I think we have a bigger problem than just circles and arcs. I would rather overhaul this system for 0.2 than having a quick fix for the next minor release...

johannes-wolf:
I do not think we have a "bigger" problem. With `place-anchors` you can already place anchors on any path position. We should just have better defaults. Maybe some "shape agnostic" anchors also help here and there, but I think the current implementation is pretty solid and versatile.

PR #213, opened 2023-09-23
Fix relative coordinate with `update: false`

SimonRask:
Before this PR, the boolean stored in the coordinate was converted to a float, which broke the logic for `update: false` of relative coordinates.

This PR:
- ensures that the boolean value is not converted to a float
- adds a regression test


johannes-wolf:
Thank you. LGTM.

johannes-wolf (approved):

PR #217, opened 2023-09-24
docs: clarify about angles

MultisampledNight:
Confused me a bit, but from experimental testing it looks like radians are being used for polar coordinates.


fenjalien:
Uh, we refer to Typst's angle type: https://typst.app/docs/reference/layout/angle/. The reason it worked in your testing is because trig functions do accept numbers but they are interpreted as radians: https://typst.app/docs/reference/foundations/calc#functions-sin.

Try using `90deg` and the likes

MultisampledNight:
Aaaaaah, that makes a lot of sense, sorry. I've updated to make this clear.

johannes-wolf (approved):

PR #221, opened 2023-09-27
Redesign internals

fenjalien:
This PR, changes the internal design of how elements are processed and handeled. Currently each element is a dictionary of functions and data that gets called/mutated at certain steps. This heavily restricts what can and can't be done without modifying `process-element` which could cause issues and bugs with other elements. It also honstely feels like spaghetti code. 

Instead, each element will ~has been~ changed to be a single function that accpets `ctx` and returns a dictionary that contains at least `ctx` but can include drawables, anchors, names etc. Ideally each element will behave the exact same so from a user point of view nothing is different. 

I will also take this opportunity to find and highlight other areas of the codebase that could do with de-spaghetti-fying/optimising/enhancement. They shouldn't be modified in this PR but I will create issues as I go.
- Marks
- Styling
- Allow content parameter in shape elements to remove frame styling in content
- Create new coordinate to remove `place-marks`

I would also like to request a freeze on merging PRs that add features as I would then have to go about implementing them again here. Also rebasing with so many changes is getting harder and harder. Important bug fixes I would be okay with handling!


src/canvas.typ
```diff
@@ -1,322 +1,116 @@
...
-  return (bounds: bounds, ctx: ctx, drawables: drawables)
-}
-
-#let canvas(length: 1cm,        /* Length of 1.0 canvas units */
-            background: none,   /* Background paint */
-            debug: false, body) = layout(ly => style(st => {
-  if body == none {
-    return []
-  }
+#let canvas(length: 1cm, debug: false, background: none, body) = layout(ly => style(st => {
```
johannes-wolf, line 13: Allow `body == none` and return nothing.

src/draw/mod.typ
```diff
@@ -0,0 +1,4 @@
+#import "grouping.typ": intersections, group, anchor, copy-anchors, place-anchors, set-ctx, get-ctx, for-each-anchor, on-layer, place-marks
```
johannes-wolf, line 1: What is draw/mod.typ for? We already have draw.typ.

src/process.typ
```diff
@@ -0,0 +1,164 @@
...
+#let element(ctx, element-func) = {
+  let bounds = none
+  // let drawables = ()
+  let element
+  let anchors = (:)
+
+
+  (ctx, ..element,) = element-func(ctx)
+  if "drawables" in element {
+    if type(element.drawables) == "dictionary" {
```
johannes-wolf, line 16: Type syntax

johannes-wolf (commented):
Look _very_ good to me! Some nitpicks…

src/draw/mod.typ
fenjalien: Its a rust-y way to do modules, its similar to `__init__.py` with python modules as well. I think moving it to `draw.typ` to avoid having to do a rename would work better

johannes-wolf:
Closed as the branch got moved to this repository.

PR #226, opened 2023-09-29
bezier: Fix extrema calculation

### LINKED ISSUE #212: arc: Bounding box calculation is wrong
The AABB calculated for arcs is incorrect.

```typ
#box(stroke: 2pt + red, cetz.canvas(length: .5cm, {
  import cetz.draw: *

  arc((0, 0), start: 45deg, delta: 270deg, radius: (5, 2), name: "c",
      anchor: "origin")
}))
```

Only happens in some cases. Cubic bezier extrema calculation fails for some curves.
Fix: Rotate curves onto the x-axis before solving dx!

johannes-wolf:
This fixes extrema calculation, which did not work in some cases. Fixes #212.


fenjalien (changes requested):
I know its not the purpose of the pr but could you please add some comments in `cubic-extrema` describing what is happening?

johannes-wolf:
> I know its not the purpose of the pr but could you please add some comments in `cubic-extrema` describing what is happening?

Will do.

PR #237, opened 2023-10-01
Improvements to `brace`

RubixDev:
This PR adds a new style key called `outer-pointiness` for the braces from the `decorations` library. It effectively controls the same thing as the existing `pointiness` but for the outer points instead. This adds the two new debug points `i` and `j` (what was `i` before is now `k`). Setting this to `0` (as is the default) provides the same behaviour as before.

I also moved documentation of the style keys from the function's doc-comment to the `manual.typ` file just like the other elements do it.

Screenshot of the updated example:
image

I am also second guessing my choice from yesterday in #227 that a pointiness of 0 is the maximum. Intuitively I would expect 1 to be the max, but then again with the angles it makes sense to have `0deg` as the max :shrug:. If you think we should change this I could quickly edit this PR or create another one, but that would of course immediately break backwards compatibility with v0.1.2.

---

Also unrelated to the PR: I noticed the example code in the manual is justified which I think doesn't look good in code snippets. And the default font inside the example canvases is also monospace.

### DIFF
--- a/manual.typ
+++ b/manual.typ
@@ -1091,6 +1091,7 @@ content((rel: (.3, .1), to: "hill.center"), text[*εїз*])
 #def-arg("pointiness", `<number> or <angle>`, default: 15deg, [How pointy the spike should be. #0deg or `0` for maximum pointiness, #90deg or `1` for minimum.])
 #def-arg("outer-pointiness", `<number> or <angle> or <auto>`, default: 0, [How pointy the outer edges should be. #0deg or `0` for maximum pointiness (allowing for a smooth transition to a straight line), #90deg or `1` for minimum. Setting this to #auto will use the value set for `pointiness`.])
 #def-arg("content-offset", `<number>`, default: .3, [Offset of the `content` anchor from the spike.])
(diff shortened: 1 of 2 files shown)


RubixDev:
I also just noticed an issue with the flip toggle: because internally it just swaps the start and end positions the final `move-to(end)` call then actually moves to the start position, which is not what users expect. I pushed a quick fix that the move-to checks for the `flip` value.

RubixDev:
> I pushed a quick fix that the move-to checks for the flip value.

Actually that's still not perfect when used as part of a `merge-path`. I'll work on a better fix in a bit.

RubixDev:
> Actually that's still not perfect when used as part of a merge-path.

Ok that's resolved. Flipping now works by negating the amplitude and pointiness values, which doesn't have these issues.

Additionally I improved the assertion messages and added a `debug-text-size` style option, so this PR is now just general polishing of yesterdays work in #227.

johannes-wolf (approved):
LGTM.

PR #245, opened 2023-10-09
tests: Allow nested tests

johannes-wolf:
Tests are now detected by looking for files named "test.typ", rather than using each directory under "$TEST_ROOT".


tests/plot/parametric/test.typ
```diff
@@ -0,0 +1,88 @@
+#set page(width: auto, height: auto)
+#import "../../../src/lib.typ": *
```
fenjalien, line 2: At this point would it be easier to use `/src/lib.typ`?

tests/plot/parametric/test.typ
```diff
@@ -0,0 +1,88 @@
...
+             domain: (0, 2 * calc.pi), fill: true)
+    plot.add((t) => (calc.cos(t * 1rad) - .5, calc.sin(t * 1rad)),
+             domain: (0, 2 * calc.pi), fill: true)
+    plot.add((t) => (calc.cos(t * 1rad), calc.sin(t * 1rad) + .5),
+             domain: (0, 2 * calc.pi), fill: true)
+    plot.add((t) => (calc.cos(t * 1rad), calc.sin(t * 1rad) - .5),
+             domain: (0, 2 * calc.pi), fill: true)
+  })
+}))
+*/
```
fenjalien, line 73: Why is this commented out?

tests/plot/parametric/test.typ
```diff
@@ -0,0 +1,88 @@
...
+
+
+/* oo */
```
fenjalien, line 76: Can this be a bit more descriptive?

fenjalien (changes requested):
I'm impressed by how little the test script needed to change. It looks like you need to update the reference images though

tests/plot/parametric/test.typ
johannes-wolf: I removed it. It gets re-added in the contour PR.

tests/plot/parametric/test.typ
johannes-wolf: I removed it. It gets re-added in the contour PR as it renders wrong in this version.

fenjalien (approved):

PR #258, opened 2023-10-17
Redesign anchor creation

fenjalien:
Elements now have to always create their own anchors instead of it happening in `process` by inferring the element's bounds. Elements must now return a function that takes a string and outputs the transformed anchor. An empty array can be passed to the function to get the list of anchor names.

I've only made these changes to `circle` as I want feedback before I go ahead. Its not perfect and could do with some tidying up as well.

I think this can also pave the way for anchor along a path and angle anchors, as they both would require a callback.


johannes-wolf:
I am sorry, I think that I have introduced all those merge conflicts by updating the `redesign-internals` branch on the current master, as it was missing one commit, unaware that this would cause problems.

fenjalien:
@johannes-wolf what do you want to do about bounding box anchors for groups? They've been removed from shapes so far.

johannes-wolf:
We shuold add them Getting a bbox of multiple elements is very important.

fenjalien:
Current notes:
- Arc angles cannot be `0deg` as they rely on border anchors for caridnal anchors. So by allowing `0deg` they have no volume/essentially a straight line which then can't be tested for intersections...
- I broke content rotation and idk how to fix it

johannes-wolf:
> Current notes:
> 
>     * Arc angles cannot be `0deg` as they rely on border anchors for caridnal anchors. So by allowing `0deg` they have no volume/essentially a straight line which then can't be tested for intersections...

Wdym with angles 0deg? Stop-Start != 0deg? I think that is fine.

>     * I broke content rotation and idk how to fix it

I fixed the rotation of content.

Some tests fail with panics.

johannes-wolf:
@fenjalien I fixed many things but group translation, which do not work at all.

johannes-wolf (changes requested):
Group translation (`anchor: "..."`) currently does nothing.

src/anchor.typ
```diff
@@ -0,0 +1,101 @@
...
+//     util.apply-transform(
+//       transform,
+//       out
+//     )
+//   } else {
+//     out
+//   }
+// }
+
+#let setup(callback, anchor-names, default: none, transform: none, name: none, offset-anchor: none) = {
```
johannes-wolf, line 33: I have to say, I do not really get this function.
How can I do calculation on (transformed) coordinates for anchor creation?
Example: Line midpoint (which got removed, why?).

src/anchor.typ
```diff
@@ -0,0 +1,101 @@
...
+        out
+      )
+    } else {
+      out
+    }
+  }
+  return (transform, calculate-anchor)
+}
+
+#let border(center, x-dist, y-dist, drawables, angle) = {
```
johannes-wolf, line 74: Suggest renaming this into `frame`.

src/anchor.typ
fenjalien: I've added some docs to this but it mainly returns another function that can be called to get the position of an anchor. Is there a situation where you would need to use transformed coordinates for anchor creation? Even so you can transform the anchors before and leave `transfrom` as `none`.

Also line never had a midpoint anchor? do you mean `center`? which it really shouldn't have. `mid` can be added with path anchors though!

src/anchor.typ
fenjalien: I suggest not as tikz refers to them as *border* anchors and this function should be used to calculate them.

fenjalien:
@johannes-wolf fixed the group anchor issue and several others. I would like to add some tests but I'm not sure what exactly.

johannes-wolf:
Some things to test:
- Group anchors with and without inner and outer transformations
- Group translation
- Anchors for all the draw commands

src/anchor.typ
johannes-wolf: Ok. Can you give an example how one could implement a custom anchor that gets calculated in the anchor callback? Something like `mid`?

johannes-wolf:
Rebased onto 0.2.0.

src/coordinate.typ
```diff
@@ -68,16 +68,22 @@
...
-    message: strfmt("Unknown element '{}' in elements {}", name, repr(ctx.nodes.keys()))
-  )
+  // Two stage anchor lookup:
+  //   1: elements => <anchor>
+  //   2: current group => anchors => <anchor>
+  let anchor = if name in ctx.nodes {
+    let node = ctx.nodes.at(name)
+    (node.anchors)(anchor)
+  } else if name in ctx.groups.last().anchors {
+    ctx.groups.last().anchors.at(name)
```
johannes-wolf, line 78: Not sure about this: This way, anchors added via `copy-anchors` can not overwrite element anchors.

PR #266, opened 2023-10-21
coordinate: Implement anchor offset syntax

johannes-wolf:
This implements an additional syntax for anchor coordinates:
  `"node.anchor[+-]*"`

Adding one or more "+" to the anchor offsets it in the direction from element center
to "node.anchor" multiplied by the number of "+" characters. The "-" character does the
same but in the opposite direction.

**Idea:** have an easy to use (short) syntax for offsetting from anchors to make placing
labels or annotations on objects.

The code currently has three TODOs:
- If the distance should be configurable: how? Via style or via an ctx setting? A reason to not using the style system is, that coordinates do not (and should not) know which element they belong to.
- Should we introduce a special "origin/center" anchor for the direction calculation?
- Should we have a default direction?

**Example:**
```typst
#cetz.canvas(length: 1cm, {
  import cetz.draw: *

  line((0,0), (4,1), name: "line")
  circle("line.end", radius: .25)
  circle("line.end+", radius: .25, stroke: green)
  circle("line.start+", radius: .25, stroke: red)
})
```

image

### DIFF
--- a/src/coordinate.typ
+++ b/src/coordinate.typ
@@ -53,28 +53,49 @@
 
 
 #let resolve-anchor(ctx, c) = {
-  // (name: <string>, anchor: <string> or <none>)
-  // "name.anchor"
+  // (name: <string>, anchor: <string> or <none>, offset: <int> or <none>)
+  // "name.anchor[+-]*"
   // "name"
 
-  let (name, anchor) = if type(c) == str {
+  // Count the number of + and - characters at the end of
+  // name, and return the name without the character and
+  // the offset as tuple.
+  let parse-offset(name) = {
+    let offset = 0
+    while name.len() > 1 {
+      if name.ends-with("+") {
+        offset += 1
+        name = name.slice(0, -1)
+      } else if name.ends-with("-") {
+        offset -= 1
+        name = name.slice(0, -1)
+      } else {
+        break
+      }
+    }
+    return (name, offset)
+  }
+
+  let (name, anchor, offset) = if type(c) == str {
     let parts = c.split(".")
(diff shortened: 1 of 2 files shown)


fenjalien:
I like the idea but we should wait for #258. It would be best to configure it through `set-style` as its supposed to be equivalent to tikz's key value setup but I think it needs another rework before that could happen...

johannes-wolf:
> I like the idea but we should wait for #258. It would be best to configure it through `set-style` as its supposed to be equivalent to tikz's key value setup but I think it needs another rework before that could happen...

If we do not have a `"center"` key on all elements, how should something like this work?
We could query multiple anchor names `("center", "mid", ..)` and use the first one that exists or fail.

fenjalien:
I think this feature would be obsolete once path anchors are implemented. I was think of allowing some kind of a shorthand similar to anchors: "line.1.1" would be beyond the end of a line.

> If we do not have a "center" key on all elements, how should something like this work?

Calling a function to get the anchor for an element allows the element to generate this when needed so it'll have to be figured out for each element. Which elements apart from line, bezier and catmull would you want this to affect?

johannes-wolf:
> I think this feature would be obsolete once path anchors are implemented. I was think of allowing some kind of a shorthand similar to anchors: "line.1.1" would be beyond the end of a line.
> 
> > If we do not have a "center" key on all elements, how should something like this work?
> 
> Calling a function to get the anchor for an element allows the element to generate this when needed so it'll have to be figured out for each element. Which elements apart from line, bezier and catmull would you want this to affect?

Should be useful for `arc`, too.

PR #267, opened 2023-10-21
feat: plot.add-fit

JamesxX:
Introducing a new plot.add-* function: lines of best fit. Currently only supports linear regression, but code is in a nice state to easily add more

### DIFF
--- a/src/lib/plot/lobf.typ
+++ b/src/lib/plot/lobf.typ
@@ -1,0 +1,64 @@
+#import "line.typ": add,
+#import "mark.typ"
+
+#let lobf-modes = (
+  "linear",
+)
+
+#let add-lobf(
+    data, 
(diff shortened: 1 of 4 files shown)


src/lib/plot/lobf.typ
```diff
@@ -0,0 +1,64 @@
+#import "line.typ": add,
+#import "mark.typ"
+
+#let lobf-modes = (
+  "linear",
+)
+
+#let add-lobf(
```
johannes-wolf, line 8: I would prefer `add-fit` and `fit.typ` as filename.

JamesxX: Committed

johannes-wolf:
Closing because `cetz-plot` will be/got moved out of `cetz`.

PR #271, opened 2023-10-21
plot: Contour Fixes

### LINKED ISSUE #270: plot: Improve contour plots
I'm really happy that we went from 0 to 60 (percent) real quick (#215). But it still has a lot of rough edges. So I figured, that I will list them here and then probably convert it to a tracking issue (of dependency sub-issues).

Notes:
- all "formulas" are assumed to be prefixed with `(x, y) => `;
- some of them are simplified (not the same as in the plot example) to simply convey the main idea.
- struck out items mean that "it is not the intended usage" (see a note below such an item).

Issues that must be fixed (high priority):
- Straight lines are offsetted.
  - <strike>No issues with conditionals (`x >= 0`, `y >= 0`, `x == y`).</strike>
    Note: boolean usage is discouraged (see [this](https://github.com/johannes-wolf/cetz/issues/270#issuecomment-1773925210) and [this](https://github.com/johannes-wolf/cetz/issues/270#issuecomment-1773934945)).
  - [x] `x` is now showing as `x = 1` (offsets by `+1` to the right) and `y` is now showing as `y = 1` (offsets by `+1` to the top).
    Note: if `z: 0` then it works correctly (default is `z: 1`, at least at the time of testing).
- Circle is filling outside the contour.
  - <strike>Correctly fills if `9 - x² + y²` (which is `9 = x² + y²` or a positive float).</strike>
    Note: use `x² + y²` and `z: 9` (for the right-hand side) instead.
  - <strike>Correctly fills if `x² + y² <= 9`.</strike>
    Note: boolean usage is discouraged (see [this](https://github.com/johannes-wolf/cetz/issues/270#issuecomment-1773925210) and [this](https://github.com/johannes-wolf/cetz/issues/270#issuecomment-1773934945)).
  - [x] <strike>Doesn't work if `x² + y² - 9` (which is `x² + y² = 9` or a negative float).</strike>
    Note: use `x² + y²` and `z: 9` (for the right-hand side) instead.
- [x] <strike>Circle contour is wrong if `bool` is returned `bool` (`x² + y² <= 9`).</strike>
  Note: boolean usage is discouraged (see [this](https://github.com/johannes-wolf/cetz/issues/270#issuecomment-1773925210) and [this](https://github.com/johannes-wolf/cetz/issues/270#issuecomment-1773934945)).
- [x] <strike>Hyperbola contour is wrong if `bool` is returned `bool` (`x ⋅ y >= 1`).</strike>
  Note: boolean usage is discouraged (see [this](https://github.com/johannes-wolf/cetz/issues/270#issuecomment-1773925210) and [this](https://github.com/johannes-wolf/cetz/issues/270#issuecomment-1773934945)).
- [x] <strike>Hyperbola contour is inverted if `1 - x ⋅ y` (negative float), but it's correct if `x ⋅ y - 1` (positive float).</strike>
Note: use `x ⋅ y` and `z: 1` (for the right-hand side) instead.

Issues that should be fixed (medium priority):
- [x] `x` can only fill to the right and `y` can only fill to the top (same with `x - y`).
  This probably can only be fixed (ideally) with `x <= 0`/`y <= 0`, i.e., changing return value from `float` to `bool`.
  Note: <strike>logical/boolean operators (conditionals) are already supported and</strike> with the right value for `x-domain`/`y-domain` the fill direction is correct. Use a new `op: "<="` (or `">="`) option and leave `x`/`y` as is.
- [ ] Contour for `x` and `y` is drawn as a C and U (sometime even O) shapes instead of straight lines (with limited length).

Features that are nice to have (low priority):
- [ ] Can't get intersection points (and their coordinates).
- [x] Multiple layers of fill don't blend/mix.
  Note: use `style: (line: (stroke: rgb("#378A467F"), fill: rgb("#378A467F").lighten(25%)))` inside the `add-contour` (see [`rgb()`](https://typst.app/docs/reference/visualize/color/#definitions-rgb)).

This issue was created after testing against the [`master#0a92100`](https://github.com/johannes-wolf/cetz/commit/0a9210095e72ec081ef9bc8a707142ef25f6aa65). The base code through which all of the above problems/missing features were discovered:

<details><summary>base.typ</summary>

```typ
#import "@local/cetz:0.1.2"

#set page(fill: black)
#set text(fill: white)

#cetz.canvas({
  import cetz.draw: *
  import cetz.plot
  let get-style(color) = (line: (stroke: color, fill: color.lighten(75%)))
  plot.plot(size: (15, 15), {
    // x >= 0
    plot.add-contour(
      (x, y) => x + 1,
      x-domain: (0, 10),
      y-domain: (-10, 10),
      fill: true,
      style: get-style(green),
    )

    // y >= 0
    plot.add-contour(
      (x, y) => y + 1,
      x-domain: (-10, 10),
      y-domain: (0, 10),
      fill: true,
      style: get-style(purple),
    )

    // hyperbola
    plot.add-contour(
      (x, y) => (x - 1) * (y - 1) - 1,
      x-domain: (-10, 10),
      y-domain: (-10, 10),
      fill: true,
      style: get-style(red),
    )

    // circle
    plot.add-contour(
      (x, y) => calc.pow((x - 1), 2) + calc.pow((y - 1), 2) - 9,
      x-domain: (-10, 10),
      y-domain: (-10, 10),
      // fill: true, // Paint outside of the circle
      style: get-style(blue),
    )

    // line
    plot.add-contour(
      (x, y) => x + 1 - y,
      x-domain: (-10, 10),
      y-domain: (-10, 10),
      style: get-style(black),
    )
  })
})
```

</details>

<details><summary>Plot screenshot</summary>

image

</details>

<details><summary>Reference screenshots from Desmos</summary>

image

image

image

</details>

---
### New issues after testing against [`redesign-internals#742d560`](https://github.com/johannes-wolf/cetz/pull/271/commits/742d560a9579ab43cc75eb3cb7af5606c1084908).

- [ ] When using `op: "=="` to plot straight lines (`x - y` + `z: 0`) nothing is plotted (the line isn't present).
- [ ] When using `x² + y²` + `op: ">="` + `z: 9` + `fill: true` it fills outside the circle **and inside** of it (although it works as expected with hyperbola). #277
- [x] With `style: (line: (stroke: color + 1em, fill: color))` (for `add-contour()`) instead of adding thickness to the contour line (border) it also increases area of the fill, including outside the borders:
  <details><summary>screenshot</summary>

  
  image

  </details>
  <details><summary>code</summary>

  ```typ
  #import "@local/cetz:0.1.2"
  
  #cetz.canvas({
    import cetz.draw: *
    import cetz.plot
    // let get-style(color) = (line: (stroke: color, fill: color.lighten(25%)))
    let get-style(color) = (line: (stroke: color + 1em, fill: color.lighten(25%)))
    plot.plot( size: (15, 15), axis-style: "school-book", {
      // x >= 0
      plot.add-contour(
        (x, y) => x,
        z: 0,
        x-domain: (0, 15),
        y-domain: (-1, 1),
        fill: true,
        style: get-style(rgb("#378A467F")),
      )
  
      // hyperbola
      plot.add-contour(
        (x, y) => (x - 1) * (y - 1),
        z: 20,
        x-domain: (-15, 15),
        y-domain: (-15, 15),
      )
    })
  })
  ```

  </details>
  Note: I forgot to remove transparency of the contour line. Use this:

  ```typ
  #let get-style(color) = {
    let (r, g, b, ..) = color.to-rgba()
    (line: (stroke: rgb(r, g, b) + 1em, fill: color.lighten(25%)))
  }
  ```
- [x] In the above scenario, the thickened contour stroke is added to the right of the RHS, instead of the center.
  Note: this is a blending effect, not a bug, see note of the task above.

johannes-wolf:
Fixes #270.

Changes:
- Fixes interpolation at domain edges.
- Adds the option to specify a comparison operator.


src/lib/plot/contour.typ
```diff
@@ -5,11 +5,17 @@
 // Find contours of a 2D array by using marching squares algorithm
 //
 // - data (array): 2D float array (rows => columns)
-// - offset (float): Value (offset) a cell must be greater or equal to to count as true
+// - offset (float): Z value (offset) of a cell compare with `op` to, to count as true
```
fenjalien, line 8: This isn't clear what offset does. Does it get added to all Z values? Is it a threshold?

src/lib/plot/contour.typ
```diff
@@ -5,11 +5,17 @@
 // Find contours of a 2D array by using marching squares algorithm
 //
 // - data (array): 2D float array (rows => columns)
```
fenjalien, line 7: This could be a bit more verbose. maybe something like "A 2D array of floats where the first index is the row and the second index is the column" or "An array of arrays of floats where the array of floats is a column and array of arrays is the row". Maybe give an example and describe what the data should represent.

src/lib/plot/contour.typ
```diff
@@ -20,15 +26,30 @@
...
   }
 
-  // Return if data is set
-  let is-set(v) = {
-    return if offset < 0 {
-      v <= offset
-    } else {
-      v >= offset
+  if type(op) != function {
+    assert(op in (auto, "<", "<=", ">", ">=", "==", "!="))
```
fenjalien, line 30: the function check should be in the assert and add an error message

src/lib/plot/contour.typ
```diff
@@ -20,15 +26,30 @@
...
+    if op == auto {
+      op = if offset < 0 { "<=" } else { ">=" }
     }
```
fenjalien, line 33: this check could go before the assert

src/lib/plot/contour.typ
```diff
@@ -5,11 +5,17 @@
...
+//       with two arguments, the z value `z1` to compare against and
+//       the z value `z2` of the data and must return a boolean: `(z1, z2) => boolean`.
 // - interpolate (bool): Enable cell interpolation for smoother lines
 // - contour-limit (int): Contour limit after which the algorithm panics
 // -> array: Array of contour point arrays
-#let find-contours(data, offset, interpolate: true, contour-limit: 50) = {
+#let find-contours(data, offset, op: auto, interpolate: true, contour-limit: 50) = {
   assert(data != none)
   assert(type(data) == array)
   assert(type(offset) in (int, float))
```
fenjalien, line 21: I'm not sure if this is relevant to the pr but error messages should be added here

src/lib/plot/contour.typ
```diff
@@ -55,6 +76,8 @@
 
   let lerp(a, b) = {
     if a == b { return a }
+    else if a == none { return 1 }
+    else if b == none { return 0 }
```
fenjalien, line 80: what happens if both are none? why can they even be none?

fenjalien (changes requested):

src/lib/plot/contour.typ
johannes-wolf: It is a threshold, I'll add that.

src/lib/plot/contour.typ
johannes-wolf: Because the algorithm look outsides the data to make the edge detection at domain limits possible. This might change with the fix for the domain edges, but is correct for filling.

fenjalien (approved):

PR #280, opened 2023-10-23
Save you some headache on `get-case`

JamesxX:
Spotted this just now, cleaned up the code to work with binary rather than decimal


johannes-wolf (approved):
:+1:

PR #282, opened 2023-10-24
Dendrogram pickup - code review 2

JamesxX:
Made most of the changes requested in the code review. There's still some more work to be done most likely, but it got a bit tricky seeing what each message was referring to given the amount of changes made

### DIFF
--- a/src/lib/chart/dendrogram.typ
+++ b/src/lib/chart/dendrogram.typ
@@ -97,7 +97,7 @@
   assert(type(height-key) in (int, str),
     message: "Invalid height-key type. Must be a integer or a string")
 
-  if ( size == auto ){
+  if size == auto {
     if mode == "vertical" {size = (auto, 1)} 
     if mode == "horizontal" {size = (1, auto)}
   }
@@ -109,13 +109,10 @@
     size.at(size-node-axis-index) = (data.len() + 2)
   }
 
-  let max-value = dendrogram-max-value-fn(data, height-key)
-  if y-max != auto { max-value = y-max }
+  let max-value = (if y-max != auto {y-max} else {calc.max(..data.map(data => data.at(height-key)))})
+  let min-value = (if y-min != auto {y-min} else {calc.min(0, ..data.map(data => data.at(height-key)))})
 
-  let min-value = dendrogram-min-value-fn(data, height-key)
-  if y-min != auto { min-value = y-min }
-
-  if (x-ticks == auto) {
+  if x-ticks == auto {
     // Pre-calculate order of leaf indices
     let x-counter = 0
     let ticks = ()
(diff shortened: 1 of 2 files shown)


src/lib/chart/dendrogram.typ
```diff
@@ -99,33 +90,53 @@
...
+  if size.at(size-node-axis-index) == auto {
+    size.at(size-node-axis-index) = (data.len() + 2)
   }
 
-  let x-ticks = x-ticks
+  let max-value = dendrogram-max-value-fn(data, height-key)
+  if y-max != auto { max-value = y-max }
+
+  let min-value = dendrogram-min-value-fn(data, height-key)
+  if y-min != auto { min-value = y-min }
```
fenjalien, line 116: These could be `if...else` statements. I think calculating the max/min in discrete functions doesn't help code readability as they aren't used anywhere else and are only one line.

src/lib/chart/dendrogram.typ
```diff
@@ -99,33 +90,53 @@
...
+  assert(type(x1-key) in (int, str),
+    message: "Invalid x1-key type. Must be a integer or a string")
+  assert(type(x2-key) in (int, str),
+    message: "Invalid x2-key type. Must be a integer or a string")
+  assert(type(height-key) in (int, str),
+    message: "Invalid height-key type. Must be a integer or a string")
+
+  if ( size == auto ){
+    if mode == "vertical" {size = (auto, 1)} 
+    if mode == "horizontal" {size = (1, auto)}
```
fenjalien, line 102: use an `if...else` statement here as mode can only be vertical or horizontal
```
if mode == "vertical" {
  ...
} else {
  ...
}
```

src/lib/chart/dendrogram.typ
```diff
@@ -219,12 +211,16 @@
   }
 
   let draw-data = (
-    if mode == "vertical" {vertical-draw-dendrogram} else
-    if mode == "horizontal" {horizontal-draw-dendrogram}
+    if mode == "vertical" {basic-draw-dendrogram} else
+    if mode == "horizontal" {basic-draw-dendrogram}
   )
```
fenjalien, line 216: this isn't needed anymore

fenjalien (changes requested):
More comments sorry 🙃 nice work though!

Could you tidy up the formatting of the if statements please? The conditional doesn't need brackets and always put the result on a new line (unless its an inline statement).
```
if conditional {
  result
}
```
Also I would prefer a single assignment for conditonal assignments. So instead of 
```
let a = 0
if b {
  a = 1
}
```
do
```
let a = if b { 1 } else { 0 }
// or
let a = if b { 
  1
} else {
  0
}
```

src/lib/chart/dendrogram.typ
JamesxX: Working on radial display now, better to leave this code in so it's there when its needed

src/lib/chart/dendrogram.typ
JamesxX: Code will be needed when radial plots are implemented

src/lib/chart/dendrogram.typ
JamesxX: done

JamesxX:
> More comments sorry 🙃 nice work though!

I appreciate the candid and constructive feedback. I've still new to Typst, and haven't written a single line of rust, so its useful to learn appropriate styles and code-smells to avoid

JamesxX:
Getting there!
image

johannes-wolf:
I'd like to implement and use polar axes for this. Maybe I'll find some time during the weekend, as I am not really happy about how the "radial" mode looks with the Cartesian axes right now.

JamesxX:
> I'd like to implement and use polar axes for this. Maybe I'll find some time during the weekend, as I am not really happy about how the "radial" mode looks with the Cartesian axes right now.

Definitely agreed. Cartesian axes are only there because of a lack of alternative at the moment. Polar axes would be great though!

PR #265, opened 2023-10-21
Box and whisker plots v0.1

JamesxX:
Began work on a chart for displaying a series of box and whiskers, in the code style of the bar charts.

PR #287, opened 2023-10-25
`plot.add-boxwhisker` - A second take on box-whisker plots

JamesxX:
In PR #265, I implemented a box and whisker chart following the style of other charts, rather than implementing it as a plot. I believe this was a mistake as it quickly became cumbersome adding more features.

Rather than implement a box and whisker chart all in one go, I've opted to make a pull request where drawing a single box and whisker is handled by a plot.

image


src/lib/plot/boxwhisker.typ
```diff
@@ -0,0 +1,58 @@
...
+            draw.line((x, start),(x, end))
+            draw.line((x - whisker-width / 2, end),(x + whisker-width / 2, end))
+        }
+
+        whisker(data.x, data.q3, data.max)
+        whisker(data.x, data.q1, data.min)
+    }
+
+    ((
+        type: "boxwhisker",
```
johannes-wolf, line 47: You should set an x-domain and y-domain, otherwise, the boxes are unable to auto-resize the plot view.

src/lib/plot/boxwhisker.typ
```diff
@@ -0,0 +1,58 @@
+#import "../../draw.typ"
+
+#let add-boxwhisker(
```
johannes-wolf, line 3: A little documentation :pleading_face:

johannes-wolf (commented):
Looks very nice! Some comments…

src/lib/plot/boxwhisker.typ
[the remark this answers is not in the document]
JamesxX: Sounds good!

JamesxX: WIP :P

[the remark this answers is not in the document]
JamesxX: Actually a thought, it would have to "guess" the x-axis values.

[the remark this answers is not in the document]
johannes-wolf: Maybe we should use tuples of the format:
`((x, (<whiskerbox-dict>)), (x, (<whiskerbox-dict>)))`

But your whiskerbox dict can even have a `x` field?

src/lib/plot/boxwhisker.typ
```diff
@@ -0,0 +1,72 @@
...
+        axes: axes,
+        bw-data: data,
+        data: (if "outliers" in data {data.outliers.map(it=>(data.x, it))} else {none}),
+        style: style,
+        plot-prepare: prepare,
+        plot-stroke: stroke,
+        mark: (if "outliers" in data {mark}),
+        mark-size: mark-size,
+        mark-style: (:),
+        x-domain: (data.x - 1, data.x + 1),
```
johannes-wolf, line 69: This should be `(data.x - calc.max(whisker-width, box-width), data.x + calc.max(whisker-width, box-width))`

src/lib/plot/boxwhisker.typ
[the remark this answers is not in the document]
JamesxX: Think tuples would work well, made sense at the time separating the X component out of the plot element

JamesxX:
image
Plot elements

image
Chart

src/lib/plot/boxwhisker.typ
```diff
@@ -0,0 +1,90 @@
...
+    assert( "x" in data, message: "Specify the x value at which to display the box and whisker")
+    assert( "min" in data, message: "Specify the q1, the minimum excluding outliers")
+    assert( "q1" in data, message: "Specify the q1, the lower quartile")
+    assert( "q2" in data, message: "Specify the q2, the median")
+    assert( "q3" in data, message: "Specify the q3, the upper quartile")
+    assert( "max" in data, message: "Specify the q1, the minimum excluding outliers")
+
+    // Calculate y-domain
+
+    let max-value = calc.max(
```
johannes-wolf, line 37: (same for `min-value): The `0` is not correct here. Use `util.max` and `util.min` from `/src/util.typ`, as they ignore `none` values.
```
let max-value = util.max(data.max, data.at("outliers"))
```

src/lib/plot/boxwhisker.typ
JamesxX: Done

JamesxX:
Last thing to do is handle passing style through the chart. I'm an absolute novice with styles so can I leave that to one of the other contributors?

johannes-wolf:
> Last thing to do is handle passing style through the chart. I'm an absolute novice with styles so can I leave that to one of the other contributors?

Yes, I can do this in a follow-up PR.

johannes-wolf (approved):

PR #288, opened 2023-10-25
tests: Test fixes

johannes-wolf:
- Remove the local package test.
- ~~Update _all_ reference images: This is actually not correct, but would make tests succeed on the 0.2.0 branch. Since some features of master are missing, the test images are not matching as of now.~~
- The test-script did not always return an != 0 exit code on failure


fenjalien:
Are you sure you want to update *all* of the tests? Most of them fail because of marks not being placed properly and they are yet to be worked on. When they do get worked on these tests will be very useful in not making a regression. So could we leave the specifically mark tests alone but the rest (which may have marks but are not the main focus) can be updated?

johannes-wolf:
> Are you sure you want to update _all_ of the tests? Most of them fail because of marks not being placed properly and they are yet to be worked on. When they do get worked on these tests will be very useful in not making a regression. So could we leave the specifically mark tests alone but the rest (which may have marks but are not the main focus) can be updated?

Then maybe update none of the tests, but only fix the test script.

fenjalien (approved):

PR #289, opened 2023-10-25
palette: Improve palettes

### LINKED ISSUE #283: Allow change in stroke style for each palette entry
https://github.com/johannes-wolf/cetz/blob/0a9210095e72ec081ef9bc8a707142ef25f6aa65/src/lib/palette.typ#L10-L17

Currently a palette can only change the fill colour, and the stroke is constant between all colours. Varying the stroke individually would be a great bonus, especially for line plots.

I need me my stroke palettes:
image

johannes-wolf:
Fixes #283.

- Adds more options to `palette.new`
- ~~By default, the stroke color gets changed to the color, too~~
- Adds various monochromatic palettes

### DIFF
--- a/src/lib/palette.typ
+++ b/src/lib/palette.typ
@@ -1,36 +1,120 @@
-/// Define a new palette
+#let base-style = (stroke: (paint: black), fill: none)
+
+/// Create a new palette based on a base style
 ///
 /// A palette is a function in the form `index -> style` that takes an
 /// index (int) and returns a canvas style dictionary. If passed the
 /// string `"len"` it must return the length of its styles.
 ///
-/// - stroke (stroke): Single stroke style.
-/// - fills (array): List of fill styles.
-/// -> function
-#let new(stroke, fills) = {
-  if type(fills) != array {fills = (fills,)}
+/// - base (style): Style dictionary to use as base style
+/// - colors (none, array): List of colors to use as stroke and fill color or none
+/// - patterns (none, array): List of stroke patterns to use or none
+/// - stroke (bool): Use color array for stroke paint
+/// - stroke-lighten (ratio): Lighten to apply to the stroke color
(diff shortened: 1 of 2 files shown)


src/lib/palette.typ
```diff
@@ -1,36 +1,120 @@
...
+/// -> function Palette function that returns a style for an index
+#let new(base: base-style, colors: (), patterns: (),
+         fill: true, fill-lighten: 0%,
+         stroke: false, stroke-lighten: 0%) = {
+  if not "stroke" in base { base.stroke = (paint: black, thickness: 1pt, dash: "solid") }
+  if not "fill" in base { base.fill = none }
+
+  let color-n = colors.len()
+  let pattern-n = patterns.len()
   (index) => {
```
johannes-wolf, line 25: Should palette functions have named arguments for some of their options (`fill:` and `stroke:`) so they can be used via `.with(...)`? @JamesxX

JamesxX: I think so, as long as it doesn't add too much overhead, it would be a really nice way to quickly touch up a palette

JamesxX:
Further, I think a palette needs an additional entry, "lerp", which takes a value between 0 and 1 and returns a style accordingly. To be used in continuous plots and heatmaps.

johannes-wolf:
> Further, I think a palette needs an additional entry, "lerp", which takes a value between 0 and 1 and returns a style accordingly. To be used in continuous plots and heatmaps.

Not sure if palettes are good for heatmaps. I would require the user to set the heat-colors separately. The code for color-lerp is already there: https://github.com/johannes-wolf/cetz/commit/99cdeb7d043e6349976f68a58e20e06f822698da (not merged, no PR yet).

johannes-wolf:
Closed in favor of updated branch.

PR #290, opened 2023-10-26
The `top` and `bottom` anchor on circle were the wrong way round

Jollywatt:
I’m on macOS and many of the tests fail (do people get imperceptible differences from system to system?), but I’m only updating the two tests that should be affected but this fix.


johannes-wolf:
Thank you! This is a known issue that got fixed in the upcoming 0.2.0 branch called `redesign-internals`. But because of the many changes, that version has not been released yet.

I wonder if we should release a tiny 0.1.3 from master which just fixes these annoying bugs? @fenjalien

johannes-wolf (approved):
LGTM.

johannes-wolf:
Which tests fail for you, @Jollywatt? Sometimes tests start failing with new Typst versions because of tiny spacing/layout changes.

fenjalien:
> I wonder if we should release a tiny 0.1.3 from master which just fixes these annoying bugs?

Hmm theres a mix of pros and cons but for the sake of using my free time well I would like to focus on finishing 0.2...

johannes-wolf:
> > I wonder if we should release a tiny 0.1.3 from master which just fixes these annoying bugs?
> 
> Hmm theres a mix of pros and cons but for the sake of using my free time well I would like to focus on finishing 0.2...

Then we don't.

PR #291, opened 2023-10-26
Angle mark miter calculation

johannes-wolf:
This re-adds the fixed miter calculation code (auto offset) for triangular marks to `mark.typ` without using it yet.


johannes-wolf:
Closed if favor of #314

PR #295, opened 2023-10-26
plot: Heatmap

johannes-wolf:
Implements a basic heatmap by drawing rects.
This is not a good solution for heatmaps with more than 15×15 tiles, as the performance is not as good for lots of rect calls.

### DIFF
--- a/src/lib/plot.typ
+++ b/src/lib/plot.typ
@@ -1,17 +1,14 @@
 // CeTZ Library for drawing plots
-#import "axes.typ"
-#import "palette.typ"
-#import "../util.typ"
-#import "../draw.typ"
+#import "/src/util.typ"
+#import "/src/draw.typ"
+#import "/src/lib/axes.typ"
+#import "/src/lib/palette.typ"
 
 #import "plot/sample.typ": sample-fn, sample-fn2
 #import "plot/line.typ": add, add-hline, add-vline
 #import "plot/contour.typ": add-contour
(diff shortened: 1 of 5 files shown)


Enivex:
I posted this feature request, which would make this vastly more performant

https://github.com/typst/typst/issues/2575

johannes-wolf:
> I posted this feature request, which would make this vastly more performant
> 
> [typst/typst#2575](https://github.com/typst/typst/issues/2575)

Thank you! I think bitmap images are the only performant option for high-res heatmaps. Even if cetz would be fast enough, the PDF would be slow to render.

Enivex:
I guess this was blocked by https://github.com/typst/typst/issues/2575 not yet being resolved?

johannes-wolf:
I've closed this by accident. Although, the current API is not that good for generating bitmaps, as it does not support raw pixel data. Having to convert to PNG/JPG first seems to be unnecessary complex.

johannes-wolf:
I've closed this by accident. Although, the current API is not that good for generating bitmaps, as it does not support raw pixel data. Having to convert to PNG/JPG first seems to be unnecessary complex.

JamesxX:
Is this something a WASM plugin could deal with?

johannes-wolf:
> Is this something a WASM plugin could deal with?

Yes. A plug-in that takes a matrix and returns a bitmap would be cool.

Issue #298, opened 2023-10-27
Allow `mark.draw-mark` to take a function for mark-size


JamesxX:
https://github.com/johannes-wolf/cetz/blob/edea3b22c6fc5b14829d09c55de8ca0c530abbfe/src/lib/plot/mark.typ#L3-L9

If `mark.draw-mark` is rewritten to allow for indexed mark sizes, bubble charts would be much easier and less costly to implement, but would mean refactoring.

johannes-wolf:
Yes, it should get rewritten. I think it should allow for indexed size + styles (stroke-color, fill-color). There should also be a more clean mechanism for plots to set marks. Maybe an extra `add-scatter`, that gets called by `add()` if the mark property is set.

JamesxX:
Styles is also a good shout, didn't see your comment by the time I made the PR

johannes-wolf:
Watch out though, the mark size is in canvas, not plot-coordinates. When using with var-sized scatter plots you would need plot coordinate scaling (which could lead to stretched marks). Maybe add a flag if the scaling should be done?

johannes-wolf:
Closed in favour of a different implementation.

PR #299, opened 2023-10-27
plot: Add Bubblechart

JamesxX:
Implements suggestion in #298. Also adds asserts for marks as the native error is absurd if you put in a wrong mark type


johannes-wolf:
Is this ready for review?

JamesxX:
Yes, I'm not about much this week so there won't be any progress on it, may as well review it

JamesxX:
I think the only thing I didn't finish implementing was scaling the ellipse in the datapoint by providing a tuple instead of a number. It's halfway implemented iirc, just needs the render side doing

PR #302, opened 2023-10-28
plot: Support reversing, rotating and locking axes

johannes-wolf:
This PR adds support for:
- Reversing an axis' direction by setting min > max
- Rotating axes by plotting with `axes: ("y", "x")`
- The orientation of custom axes can be set via `<axis>-horizontal: <bool>`
- Axes can be locked to keep aspect ratio of 1:1 using `<axis>-equal: "<axis>"` (like matplotlib `"equal"`)

Setting up axes has been moved to `util.typ` to declutter `plot.plot` a bit.


src/lib/axes.typ
```diff
@@ -55,6 +55,8 @@
 
 // Format a tick value
 #let format-tick-value(value, tic-options) = {
+  if value == -0 { value = 0 }
```
fenjalien, line 58: Typst doesn't have negative zeros?

src/lib/axes.typ
```diff
@@ -128,9 +130,10 @@
   let l = ()
   if ticks != none {
     let major-tick-values = ()
-    if "step" in ticks and ticks.step != none {
+    if "step" in ticks and ticks.step != none and ticks.step != 0 {
       assert(ticks.step >= 0,
              message: "Axis tick step must be positive")
```
fenjalien, line 135: Why not panic if `ticks.step` is 0? Also to check use `ticks.step not in (none, 0)`

src/lib/plot.typ
```diff
@@ -25,6 +26,12 @@
   return default-plot-style(i)
 }
 
+// Get the default axis orientation
+// depending on the axis name
+#let get-default-axis-horizontal(name) = {
+  return name.starts-with("x") or name.starts-with("X")
```
fenjalien, line 32: Use [lower](https://typst.app/docs/reference/text/lower/)

src/lib/plot.typ
```diff
@@ -167,29 +210,33 @@
...
-
     axis.label = get-axis-option(name, "label", $#name$)
+
+    // Configure axis bounds
     axis.min = get-axis-option(name, "min", axis.min)
     axis.max = get-axis-option(name, "max", axis.max)
 
+    assert(axis.min != auto and axis.min != none and
+           axis.max != auto and axis.max != none,
+      message: "Axis min and max must be set.")
```
fenjalien, line 221: Use `... not in (...)` `axis.min not in (auto, none) ...`

src/lib/plot.typ
```diff
@@ -96,6 +103,12 @@
 ///                    `x-min: 0`.
 ///                    - min (int): Axis minimum
 ///                    - max (int): Axis maximum
+///                    - horizontal (bool): Axis orientation; note that each
+///                      plot must use one vertical and one horizontal axis!
+///                      The default value for this parameter is guessed: Axes
+///                      starting with "x" are considered horizontal by default.
+///                      This does not affect the side the ticks of the axis are
+///                      drawn, but only the drawing direction.
```
fenjalien, line 111: I don't think this option is the best way to specify which axis should be horizontal. Its the having to check they are mutually exculsive part that does it. I think it shouldn't be so reliant on the axis setting.

Is this redundent given you have swapped axes in `plot.add` for the tests? Why not just document that properly? If not a test for `horizontal` should be added.

fenjalien (changes requested):

src/lib/axes.typ
johannes-wolf: This code worked, though. Otherwise, we get negative zeros when doing the string conversion.

src/lib/plot.typ
johannes-wolf: Since axes are created on demand and the orientation is not bound to their use, we need an option to specify the orientation. You can also use axes in any combination you want.
This is not redundant when using custom axes, e.g. `axes: ("abc", "my-axis2")`.

src/lib/axes.typ
johannes-wolf: Typst actually has negative zero floats, so this test makes sense.

johannes-wolf:
Plot `plot.typ` needs a bit more cleanup, but I will do that in a later ticket.

src/lib/axes.typ
```diff
@@ -55,6 +55,10 @@
 
 // Format a tick value
 #let format-tick-value(value, tic-options) = {
+  // Without it we get negative zero in conversion
+  // to content! Typst has negative zero floats.
+  if value == 0 { value = 0 }
```
fenjalien, line 60: wait shouldn't this be checking for `-0.0` then? Maybe add a test for this?

fenjalien (approved):

src/lib/axes.typ
johannes-wolf: It does: `-0 == 0`… there is no way to detect `-0`, because by IEE definition it must compare equal to 0 (you can only know by checking the sign bit). I was complaining on Discord that Typst should never format `-0` as `0`.

PR #307, opened 2023-10-30
matrix: Unroll matrix-vector product

johannes-wolf:
This change gives about -100 ms for a line plot with 1000 data points.
Since matrix-vector multiplication is done a lot, this specialization should be used. 
We actually never user non 4x4 matrices, so we could think about not supporting other matrix sizes in general.

### DIFF
--- a/src/matrix.typ
+++ b/src/matrix.typ
@@ -123,6 +123,23 @@
   )
 }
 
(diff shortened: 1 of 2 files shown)


fenjalien (approved):
Huh, nice work. I would like to keep the general version just in case. If `mul-vec` isn't used anywhere else its all good.

PR #310, opened 2023-10-30
plot: Legend support

johannes-wolf:
This adds legend support to plots. The legends displays as soon as at least one plot command
has `label` set.


src/lib/plot.typ
```diff
@@ -225,20 +238,60 @@
     }
   }
 
-  // Prepare
```
johannes-wolf, line 228: All changed because of indentation: I moved the plot including axes into an inner group to be able to query it's bounds.

johannes-wolf:
@JuliDi anything missing from this PR?
See the test reference pictures for how it looks: https://github.com/johannes-wolf/cetz/pull/310/files#diff-5c6c853bb2dbed33c1926068b47b643b43b0d2eaaa6802fdb4674603ffd1437d

JuliDi:
Looks good to me. I haven't had the chance to give it a try yet, though.

fenjalien:
I would rather having the padding option to be in styling, ideally all named arguments that could be shared between elements should be in styles. Especially since its supposed to be the alternative to tikz's key value pairs.

Could we allow `x`, `y` and `rest` options for padding? It would make it inline with Typst's padding: https://typst.app/docs/reference/layout/pad/ so we can reference that instead of CSS.

Also note that tikz allows padding inside and out of a node (https://tikz.dev/tikz-shapes#sec-17.2.3). I don't know how relevant that is to groups but it might be to content.

johannes-wolf:
> I would rather having the padding option to be in styling, ideally all named arguments that could be shared between elements should be in styles. Especially since its supposed to be the alternative to tikz's key value pairs.
> 
> Could we allow `x`, `y` and `rest` options for padding? It would make it inline with Typst's padding: https://typst.app/docs/reference/layout/pad/ so we can reference that instead of CSS.
> 
> Also note that tikz allows padding inside and out of a node (https://tikz.dev/tikz-shapes#sec-17.2.3). I don't know how relevant that is to groups but it might be to content.

Updated it. Supporting typst padding dictionaries by applying the "fallbacks" (x, y, rest) onto the dictionary in `util.as-padding-dict(...)`. Also added a style for `group`.

src/draw/grouping.typ
```diff
@@ -70,7 +78,8 @@
       anchor => {
         let anchors = (:)
         if add-bbox-anchors {
-          let bounds = bounds
+          let bounds = aabb.padded(bounds, util.as-padding-dict(style.padding))
```
fenjalien, line 81: this could be moved outside the anchor function

fenjalien (approved):

johannes-wolf:
> I would rather having the padding option to be in styling, ideally all named arguments that could be shared between elements should be in styles. Especially since its supposed to be the alternative to tikz's key value pairs.
> 
> Could we allow `x`, `y` and `rest` options for padding? It would make it inline with Typst's padding: https://typst.app/docs/reference/layout/pad/ so we can reference that instead of CSS.
> 
> Also note that tikz allows padding inside and out of a node (https://tikz.dev/tikz-shapes#sec-17.2.3). I don't know how relevant that is to groups but it might be to content.

We might want to support margin, too. Padding affects the (border) anchors, margin would not.
Maybe in a later PR.

PR #313, opened 2023-10-31
intersect: Add precise line-bezier intersection

johannes-wolf:
Use this for the border anchor calculation. Path/path intersection still defaults to the sampling based linear intersection.


src/intersection.typ
```diff
@@ -36,6 +36,49 @@
...
+  }
+  return pts
+}
+
+/// Check for line-path intersection in 2D
+///
+/// - la (vector): Line start
+/// - lb (vector): Line end
+/// - path (path): Path
+/// - samples (int): Number of samples per curve
```
fenjalien, line 62: samples doesn't exist

src/intersection.typ
```diff
@@ -36,6 +36,49 @@
   }
 }
 
+// Check for line-cubic bezier intersection
+#let line-cubic(la, lb, s, e, c1, c2) = {
+  import "/src/bezier.typ"
+  return bezier.line-cubic-intersections(la, lb, s, e, c1, c2)
+}
```
fenjalien, line 43: Why not rename this function on import?
```
#import "/src/bezier.typ": line-cubic-intersections as line-cubic

fenjalien (approved):
this is cool, nice work. Could `path-path` be improved by using `line-path` somehow?

johannes-wolf:
Changed. Also found a little bug in the intersection code :)

PR #314, opened 2023-11-04
mark: Mark auto offsetting 

### LINKED ISSUE #218: draw: Update mark positioning for bezier, bezier-through and place-marks
Fix `bezier.shorten(...)` which is currently not working correctly.

### LINKED ISSUE #239: Marks on Curves are Broken
Fails to build in 1.2:
```
#cetz.canvas({
  import cetz.draw: *
  line((0, 0), (3, 0), mark: (end: ">", angle: 90deg))
  bezier((0, 0), (2, 0), (1, 1), mark: (end: ">", angle: 90deg))
})
```

### LINKED ISSUE #247: Mark Lines have Bad Anchors
See: #246. The anchors seem to get moved with the mark, which is wrong.

johannes-wolf:
This is work in progress.

**Changes:** Remove `mark.size` in favor of `mark.width` and `mark.length` + `mark.scale`.
- `width`: The width of the mark measured at the base (or mid)
- `length`: The length of the mark on the path
- `scale`: A factor that scales `width` and `length` to make it easy changing mark size without having to (re-) calculate ratio
- `inset`: (filled triangles and harpoons) Sets the base inset to create an arrow-head like mark
- `sep`: Extra space between multiple marks
- _bezier & catmull_ `shorten`: Bézier shorting mode "LINEAR" or "CURVED", use linear or curved offsetting
- _bezier & catmull_ `flex`: If true, the mark points in the direction of the secant between its base and tip (default: `true`)
- You can now place multiple marks using `(">", ">")` arrays instead of a single symbol string.
- New `left-harpoon` and `right-harpoon` marks
- Circle marks (`"o"`) now use their border as tip (before it was the center)
- Catmull curves now support marks (limitation: only the first segment is used)
- Arcs now allso support marks :)

This tries to mirror TikZ main mark size related attributes.

@fenjalien You wanted to change how marks work, what is your plan?

Fixes #247
Fixes #239
Fixes #218


src/lib/angle.typ
fenjalien, line 1: This probably isn't the right pr to fix this but why is angle done with drawables and not a group with shapes?

src/mark.typ
```diff
@@ -1,5 +1,387 @@
...
+    }
+  }
+
+  if join == "bevel" {
+    return calc.sin(angle / 2) * width / 2 * sign
+  } else {
+    return width / 2 * sign
+  }
+
+  return 0
```
fenjalien, line 51: Unreachable

src/mark.typ
fenjalien, line 1: This file would benefit from a description of all the terms used, similar to what tikz does in its arrows page https://tikz.dev/tikz-arrows

Assuming the arrow is pointing directly to the right `->`
- `length`: the horizontal distance between the tip of the arrow and its "back-end"
- `width`: the vertical distance between the arrow's highest and lowest point
- `inset`: The horizontal distance between the back-end and how far something is set into the arrow head
- `offset`: ???

Add more if needed

src/mark.typ
```diff
@@ -1,5 +1,387 @@
...
+    return _triangular-mark-offset(ctx, style.width, style.length / 2, symbol, style)
+  } else {
+    let width = line(stroke: style.stroke).stroke.thickness
+    if width == auto { width = 1pt }
+    if type(width) == length { width /= ctx.length }
+
+    return -width / 2
+  }
+
+  return 0
```
fenjalien, line 73: unreachable

src/mark.typ
```diff
@@ -1,5 +1,387 @@
...
+  } else if symbol in ("left-harpoon", "right-harpoon") {
+    return _triangular-mark-offset(ctx, style.width / 2, style.length, symbol, style)
+  } else if symbol == "<>" {
+    return _triangular-mark-offset(ctx, style.width, style.length / 2, symbol, style)
+  } else {
+    let width = line(stroke: style.stroke).stroke.thickness
+    if width == auto { width = 1pt }
+    if type(width) == length { width /= ctx.length }
+
+    return -width / 2
```
fenjalien, line 70: what's happening here?

src/mark.typ
```diff
@@ -1,5 +1,387 @@
...
+  }
+
+  return 0
+}
+
+// Get mark symbol mid length, that is the length from the tip to the mid-point
+// of its base. For triangular shaped marks, that is the length minus the inset.
```
fenjalien, line 77: so the distance between the inset and the tip?

src/mark.typ
```diff
@@ -1,5 +1,387 @@
...
+  }
+
+  return length
+}
+
+/// Place marks along line of points
+///
+/// - ctx (context): Context
+/// - pts (array): Array of vectors
+/// - marks (style): Mark style dictionary
```
fenjalien, line 93: call this style

src/path-util.typ
```diff
@@ -1,4 +1,6 @@
 // This file contains utility functions for path calculation
+#let typst-length = length
```
fenjalien, line 2: Can this go in `util.typ`?

fenjalien (changes requested):
Wow I cannot believe how good this is :D

- I don't understand the purpose of `scale`, could you link its tikz alternative?
- Would it be possible to generalise the mark placement functions to work on path drawables? Then they can just modify the drawables to shorten them?

This is a great start and covers a lot of what I wanted to address. However I think there could be many more options/stuff brought over. This PR is already a good size though so it should go into a new one. Do you want me to make an issue detailing all the options we could have?

johannes-wolf:
> Wow I cannot believe how good this is :D
> 
>     * I don't understand the purpose of `scale`, could you link its tikz alternative?

It is here: https://tikz.dev/tikz-arrows#sec-16.3.2 - A factor that gets applied on all size-related (length, width, inset) attributes.

>     * Would it be possible to generalise the mark placement functions to work on path drawables? Then they can just modify the drawables to shorten them?

Yes, I will add that.

> This is a great start and covers a lot of what I wanted to address. However I think there could be many more options/stuff brought over. This PR is already a good size though so it should go into a new one. Do you want me to make an issue detailing all the options we could have?

Yes, that would be nice.

src/lib/angle.typ
johannes-wolf: I will change that in a separate PR.

src/bezier.typ
```diff
@@ -236,55 +236,118 @@
...
+    s = vector.add(s, offset)
+    c1 = vector.add(c1, offset)
+  } else {
+    e = vector.add(e, offset)
+    c2 = vector.add(c2, offset)
+  }
+  return (s, e, c1, c2)
+}
+
+/// Approximate t for a givend distance d.
```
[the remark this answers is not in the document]
johannes-wolf: The bezier value in the interval [0,1]. I could not find if it has a special name.

src/bezier.typ
```diff
@@ -236,55 +236,118 @@
...
+      let dist = vector.dist(last, curr)
+      travel += dist
+      if travel >= d {
+        split-t = t - 1/samples + d / (travel * samples)
         break
       }
       last = curr
     }
+  } else {
+    let (e, s, c2, c1) = cubic-shorten(e, s, c2, c1, -d, samples: samples)
```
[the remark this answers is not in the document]
johannes-wolf: I added a comment. It swaps the curve to run the function backwards without needing a second implementation.

src/mark.typ
johannes-wolf: It is described in `styles.typ`.

johannes-wolf:
I applied the review annotations.

johannes-wolf:
+ Fixed `vector.coss` (oops) and marks pointing to +/- z.

src/matrix.typ
```diff
@@ -83,26 +82,27 @@
...
+///
+/// Calculates the product of the three rotation matrices
+/// R = Rz(z) Ry(y) Rx(x)
+///
+/// - x (angle): Rotation about x
+/// - y (angle): Rotation about y
+/// - z (angle): Rotation about z
+/// -> matrix
 #let transform-rotate-xyz(x, y, z) = {
-  // let (pi, cos, sin) = (calc.pi, calc.cos, calc.sin)
```
johannes-wolf, line 95: This was wrong.

src/bezier.typ
```diff
@@ -236,55 +236,126 @@
...
+    }
+    last = pt
+  }
+  return d
+}
+
+/// Shorten curve by offsetting s and c1 or e and c2
+/// by distance d. If d is positive the curve gets shortened
+/// from s on, if d is negative it gets shortened form its
+/// end e.
```
fenjalien, line 261: "Shorten *the* curve"
I still don't think this is very clear, if d is positive does s stay the same and e gets closer to s? in which direction does it go as well, does e go along the curve or directly towards s?

johannes-wolf: It goes along the curve. I'll improve the comment.

johannes-wolf: I've updated the comment. The "linear" function does not move s/e on the curve, the non-linear function keeps the curvature by moving both points on the curve.

fenjalien (approved):

Issue #318, opened 2023-11-05
Hobby curves


Enivex:
http://www.jakelow.com/blog/hobby-curves

These would be a great addition to cetz. I use them all the time with tikz (https://ctan.org/pkg/hobby?lang=en) because they just look so nice.

PR #319, opened 2023-11-06
Hobby Curve Implementation

johannes-wolf:
Implements #318, thanks to @Enivex for the implementation!

This implements Hobby curves using the same algorithm mentioned in the ticket.
Plus, simplifies and improves catmull + hobby curve mark placement.


Enivex:
The hobby algorithm can produce nice smooth closed curves by the way. But the author of the blog post may not have implemented them. In any case it's described both in the TUG paper they based their code on, and the original Hobby paper.

johannes-wolf:
Ok, I will take a look at that.

johannes-wolf:
Could not get good results :(
@fenjalien, I would like to merge this in my other branch?
I will open a ticket to fix the closing and disallow it for now.

PR #324, opened 2023-11-10
Add transform function + Improve rotate

johannes-wolf:
- Adds a new `transform(matrix)` function for setting or resetting the transformation matrix
- Changes `rotate` to take a single angle or a set of named arguments of one of the lists: x, y, z or yaw, pitch, roll to create a rotation matrix


src/draw/transformations.typ
```diff
@@ -3,24 +3,57 @@
...
-      
-      return (ctx: ctx)
-    },
-  )
+/// Sets the transformation matrix
+///
+/// - mat (none,matrix): The 4x4 transformation matrix to set. If `none` is
+///   passed, the transformation matrix is set to the identity matrix (
+///   `matrix.ident()`).
+#let transform(mat) = {
```
fenjalien, line 11: I would prefer the function name `set-transformation` given it sets the current transformation matrix. `transform` suggests that it applies the given matrix to the current transformation

src/draw/transformations.typ
```diff
@@ -3,24 +3,57 @@
...
+  (ctx => {
+    assert(mat == none or type(mat) == array,
+      message: "Transformtion matrix must be none (ident) or of type array")
+    ctx.transform = if mat != none {
+      assert.eq(mat.len(), 4,
```
fenjalien, line 16: this should be checked outside the returned function

src/draw/transformations.typ
```diff
@@ -3,24 +3,57 @@
...
+  assert(angles.pos().len() == 1 or angles.named().len() > 0,
+    message: "Rotate takes a single z-angle or angles (x, y, z or yaw, pitch, roll) as named arguments")
+
+  let named = angles.named()
+  let names = named.keys()
+
+  let mat = if angles.pos().len() == 1 {
+    matrix.transform-rotate-z(angles.pos().at(0))
+  } else if names.any(n => n in ("x", "y", "z")) {
+    assert(names.all(n => n in ("x", "y", "z")),
```
fenjalien, line 36: Couldn't you just use `all` in the if statement? Then if the names are malformed you panic in the `else`? Also make sure to output what the given names were

fenjalien (changes requested):

johannes-wolf:
@fenjalien Applied all your remarks :)

src/draw/transformations.typ
```diff
@@ -3,24 +3,56 @@
...
+
+  (ctx => {
+    ctx.transform = mat
+    return (ctx: ctx)
+  },)
+}
+
+#let rotate(..angles) = {
+  assert(angles.pos().len() == 1 or angles.named().len() > 0,
+    message: "Rotate takes a single z-angle or angles (x, y, z or yaw, pitch, roll) as named arguments")
```
fenjalien, line 31: output incorrect value

johannes-wolf:
Updated the ref-images (rotation is now CCW) and added the bad values to the error messages.

fenjalien (approved):

PR #325, opened 2023-11-11
Update test ref images

johannes-wolf:
Update test reference images after mark offset reimplementation.

- Various bug fixes (marks, axes, padding)
- Updates all reference images, fixed or moved some tests

After merging this, tests should be useful again.


src/mark.typ
```diff
@@ -8,7 +8,7 @@
 // Calculate offset for a triangular mark (triangle, harpoon, ..)
 #let _triangular-mark-offset(ctx, mark-width, mark-length, symbol, style) = {
   let revert = symbol == "<"
-  let sign = if revert { 0 } else { -1 }
+  let sign = if revert { 1 } else { -1 }
```
johannes-wolf, line 11: Oops

fenjalien (approved):
As long as you are sure that there are no more differences that we won't miss its all good!

johannes-wolf:
> As long as you are sure that there are no more differences that we won't miss its all good!

I checked all test images.

PR #326, opened 2023-11-11
draw: Do not transform content anchors

### LINKED ISSUE #256: Content vertices get rotated/scaled by the current transformation matrix
Content behaves differently than all other elements in that it should not get rotated or scaled with the canvas. This is true for the content itself, but not so for the vertices of anchors, bounds, and the frame.

Reproduce:
```
#import "@preview/cetz:0.1.2"

#cetz.canvas({
  import cetz.draw: *
  rotate(45deg)
  content((0,0), [Test Text], frame: "rect")
})
```

grafik

johannes-wolf:
Content must not get transformed.
The only anchor that gets transformed is the `center` anchor. All other anchors must be calculated relative to that one.

Fixes #256.

Also fixes content bounding box calculation. Calculating the width/height from north-south and east-west is not enough. Instead, use the already calculated anchors at all four corners to calculate the rotated width/height.


tests/content/transform/test.typ
fenjalien, line 1: could we have a reference point at `(0,0)` before the transformation for each test? also make sure to test the `anchor` parameter

johannes-wolf: :+1: Added tests for all anchors + showing the (0,0) pre and post transformation.

johannes-wolf:
Ok… typst 0.9.0 crashes. My local build (main branch) does not.

PR #330, opened 2023-11-12
Update Angle Library

### LINKED ISSUE #327: angle: Make `angle` use the non-internal API
There is no need anymore for angle to use the internal element API.

johannes-wolf:
Use non-internal API for drawing angles.

Fixes #327.


tests/angle/ref.png
fenjalien, line 1: The label in the top left is missing?

johannes-wolf: Yes, this is on purpose. I think it makes no sense displaying a label for a 0 degree angle.

johannes-wolf: … but I re-added it.

Issue #331, opened 2023-11-12
Mark Improvements List


fenjalien:
Here are my suggestions of features/improvements to make for parity with tikz (from #314). Some of them we already have (not sure exactly which), some are implemented natively by typst.

## Styling:
- [ ] [length](https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/length)
    - [ ] line width factor
    - [ ] outer factor
- [ ] [width](https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/width)
    - [ ] line width factor
    - [ ] outer factor
- [ ] [width'](https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/width')
- [ ] [inset](https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/inset)
    - [ ] line width

## Configuration options
- [ ] line width factor and outer factor options for length/width/inset/angle options (https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/length)
- [ ] note that length of an arrow is calculated assuming it has a miter line cap even if the cap is different (see https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/length "The Exact Length")
- [ ] angle (https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/angle)
- [ ] length factor and line width factor options for width/inset (https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/width')
- [ ] scale length/width (https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/scale:length)
- [ ] arc (https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/arc)
- [ ] slant (https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/slant)
- [ ] Reversing tips https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/reversed
- [ ] Halving tips (https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/harpoon) (including which side to half)
- [x] bending and flexing (https://tikz.dev/tikz-arrows#sec-16.3.8) (haven't properly checked this but i think we have everything)
- [ ] per arrow config

- Note that coloring and line styling is skipped as they are handled by fill and stroke options

## tip specifications
- [ ] line width factor and outer factor for separations (https://tikz.dev/tikz-arrows#pgf./pgf/arrow:keys/sep)
[...]

PR #333, opened 2023-11-12
mark: Some Mark Features

johannes-wolf:
- Added many additional arrow styles: "->", "<-", "~>", "<~", "]", "[", "[]", "x", "+", "*", "hook", "hook-right"
- Added slanting for all arrows via style key `slant: <angle>`
- Added thickness relative scale/length/width/inset
- Added `shorten-to` that control how a path gets shortened if multiple marks are placed
- Fixed shortening: Beziers now "snap" to the correct position to prevent gaps because of low precision arc-length calculation
- Moved all mark drawing to `mark.typ`
- Fixes `grid` being to wide
- Removed harpoons: they should be reimplemented in a follow-up
- The default mark ">" changed to another style; the old one is now "->"
- A new `reverse: <bool>` style to reverse all marks
- A new `flip: <bool>` style to flip all marks

Implements some parts of #331.

Some styling options:
image

The new default tip: 
image


johannes-wolf:
Closing if favor of #416.

PR #334, opened 2023-11-16
fix: error in documentation

alxrey:


johannes-wolf:
Thank you for your PR. With 0.2.0 (the current dev branch for the next release) there are lots of changes to the documentation, so this fix does not apply anymore.

Issue #136, opened 2023-08-20
Question: linear programming solver and contouring


tmbb:
I'd like to discuss two features which are probably not realistically possible to implement in Typst itself but which may be possible with some kind of rust bindings (which I believe are impossible with Typsts the way it works now).

## Countours of surfaces

While I was implementing my own (very simple and totally not extensible) plotting library in Typst, I noticed that most things were very easy as long as I restricted myself to basic shapes, such as circles, rectangles and paths. This allows one to build plots such as pie charts, bar charts, boxplots, filled plots, Kaplan-Meier survival curves, etc. However, sometimes one might want to represent a smothed version of a 3D surface projected into 2D or a density plot, such as this (taken from the Matplotlib docs):
 
![image](https://github.com/johannes-wolf/typst-canvas/assets/2668874/8b9709fd-b81f-455f-bea3-367c3d5c72e8)

The colors themselves can be either discrete (using contours) or smooth (with some kind of linear interpolation between contours). The main algorithm used for those contours is the marching squares algorithm, which is slow and hard to implement in a scripting language like Typst. Fortunately, there seem to be very good implementations in rust for contour drawing.

## Linear Programming

Something else which is kinda hard to do and which causes problems in many plotting libraries such as Matplotlib is mixing variable and fixed lengths. For examle, suppose you have a boxplot annotated with p values from comparisons beween groups (drawn in matplotlib):

![bitmap](https://github.com/johannes-wolf/typst-canvas/assets/2668874/47f3b7b3-19d3-4b6d-b6c5-647f8e8f4378)

Because matplotlib doesn't handle fixed and variable dimensions inside plots very well, I've had to fix the limits through trial and error. Solving linear constraints for placement of items inside the plot is actually very easy, and doesn't require a real linear programming solver such as the simplex algorithm. However, if you want to take into accound the labels, I believe (but I haven't proved) that you start having to solve general linear programming for it to work. The asymptote library (a [C++ program]() to draw graphics which uses TeX for text rendering actually implements the simplex method to handle these constraints). I believe (but again, I haven't proved) that solving "general" linear programming is required to determine the margins of a plot in more complex layouts such as these:

![bitmap](https://github.com/johannes-wolf/typst-canvas/assets/2668874/ab0fc570-e345-4b60-a3dc-d48cd76b2dd6)

[...]

PR #336, opened 2023-11-18
Plot Annotations

johannes-wolf:
Implements #136 (in plot coordinates).

This PR adds `add-annotation` to allow drawing in plots. The plot also gets resized to the drawing if it exceeds the axis bounds.


src/lib/plot/annotation.typ
```diff
@@ -0,0 +1,64 @@
...
+///
+/// An annotation is a sub-canvas that uses the plots coordinates specified
+/// by its x and y axis.
+///
+/// Bounds calculation is done naively, therefore fixed size content _can_ grow
+/// out of the plot. You can adjust the padding manually to adjust for that. The
+/// feature of solving the correct bounds for fixed size elements might be added
+/// in the future.
+///
+/// - body (drawable): CeTZ Draw commands
```
fenjalien, line 17: Use `elements`

src/lib/plot/annotation.typ
```diff
@@ -0,0 +1,64 @@
...
+/// - axes (axes): X and Y axis names
+/// - resize (bool): If true, the plots axes get adjusted to contain the annotation
+/// - padding (none,number,dictionary): Annotation padding that is used for axis
+///   adjustment
+/// - background (bool): If true, the annotation is drawn behind all plots, in the background.
+///   If false, the annotation is drawn above all plots.
```
fenjalien, line 23: Could these be put into a style?

fenjalien (commented):
Looks good but an example would be nice

src/lib/plot/annotation.typ
johannes-wolf: Does not really make sense imho., as all three depend on the plot/data and are in plot units (padding).

PR #343, opened 2023-11-22
styles: Merge Strokes

johannes-wolf:
This supports merging style strokes, which is _important_ to be able to set e.g., the cap at the root style without having to repeat it over and over when changing color, etc.

With this change, you can set a different cap style for marks, while still having the color inherited from their parent.


fenjalien:
I am against this change as styling is intended to only merge dictionaries, not values that can behave like dictionaries. What should happen if a style value could be a color, length or dictionary but not a stroke? If the user needs styles to merge then they should give the stroke as a dictionary. That said I wouldn't be against converting a stroke into a dictionary when inserting it.

johannes-wolf:
> I am against this change as styling is intended to only merge dictionaries, not values that can behave like dictionaries. What should happen if a style value could be a color, length or dictionary but not a stroke? 

That would be a quite strange API tbh., but you are right.

> If the user needs styles to merge then they should give the stroke as a dictionary. That said I wouldn't be against converting a stroke into a dictionary when inserting it.

Ok, I will change that. This "magic" should happen in `set-style` then, I think. We should change all our base stroke styles to dictionaries.

johannes-wolf:
@fenjalien All stroke merging hacks are gone, but I've changed the default stroke to a dictionary, so merging is possible as long as you provide dicts to `set-style`.

Issue #180, opened 2023-09-13
Support for rounded rect


Zheoni:
Self explanatory title. I think rect could benefit from a `radius` parameter to make the corners round.

MultisampledNight:
For what it's worth, this is the polyfill I'm using for now, which requires `a` and `b` to be resolved though. One can specify `bendness: none` to use an actual arc instead of the squircle-ish bezier curve.

```typst
#import cetz.draw: *

#let vec-add(a, b) = {
  a.zip(b).map(combi => combi.at(0) + combi.at(1))
}
#let vec-sub(a, b) = {
  a.zip(b).map(combi => combi.at(0) - combi.at(1))
}
#let vec-div-scalar(a, b) = {
  a.map(a => a / b)
}

#let rounded-rect(from, to, bendness: 0, radius: 1) = group({
  let size = vec-div-scalar(vec-sub(to, from), 2)
  set-origin(vec-add(from, size))

  let outline = merge-path({
    for _ in range(4) {
      let x-max = vec-sub(size, (radius, 0))
      let y-max = vec-sub(size, (0, radius))

      line(
        (horizontal: (0, 0), vertical: size),
        x-max,
      )
      if bendness == none {
        // round it
        arc(x-max, start: 90deg, stop: 0deg)
      } else {
        // squircle it
        bezier(
          x-max,
          y-max,
          vec-add(size, (bendness, bendness)),
        )
      }
      line(
        y-max,
        (vertical: (0, 0), horizontal: size)
      )

      rotate(90deg)
      size = (size.at(1), size.at(0))
    }
  })

  outline
})
```

PR #350, opened 2023-11-24
shapes: Radius Support for Rect

johannes-wolf:
Implements #180. You can configure each x and y radius per corner (8 radii in total).

Also fixes some issues with path interpolation in path-util. Path utils `point-on-path` now uses ratios for relative distances, while floats represent absolute distances on the path.
I also marked multiple functions from `path-util` private by prefixing them with an underscore and also removed some dead code.

This also changes `rect` into only accepting points with the same z value.


MultisampledNight:
Hi, this works really neat, thank you :sparkles:! The only issue there is seems to be that rectangle radii larger than the size have a slightly bugged radius:

graphics

<details>

<summary>Generation code</summary>

```typst
#import "@local/cetz:0.2.0"
#import cetz.draw: *

#let examples(..codes) = group({
  for code in codes.pos() {
    eval("import draw: *\n" + code, scope: (draw: cetz.draw))
    content((0, -1.5), raw(lang: "typc", code))
    set-origin((6.5, 0))
  }
})

#set page(fill: white, width: auto, height: auto, margin: 0.5cm)
#cetz.canvas({
  examples(
    "rect((0, 0), (0, 0), radius: 1)",
    "rect((0, 0), (1, 0), radius: 1)",
    "rect((0, -1), (0.5, 1), radius: 1)",
  )
})
```

</details>

Maybe the radius could be limited to half the length of the shortest rectangle edge?

johannes-wolf:
@MultisampledNight Fixed those. The radius per edge gets now clamped between 0 and half the rects width/height.

johannes-wolf:
Something is still off… Path utils path calculations fail with "unoptimized" paths (with multiple line segments) rect now generates.

MultisampledNight:
Another small thing I noticed: Is it intended that using absolute lengths is unsupported?

```typ
#cetz.canvas({
  let rect = rect.with(radius: 7.5cm)
  rect((0, 0), (4, 4))
})
```

I'm checked out on 5d80997.

<details>
<summary>Error as emitted by the compilation</summary>

```typ
error: cannot compare float and length
    ┌─ @local/cetz:0.2.0/src/util.typ:298:20
    │
298 │       } else { v }, size.at(i) / 2))
    │                     ^^^^^^^^^^^^^^

help: error occurred in this call of function `as-corner-radius-dict`
    ┌─ @local/cetz:0.2.0/src/draw/shapes.typ:695:45
    │
695 │            south-west: sw, south-east: se) = util.as-corner-radius-dict(style.radius, size)
    │                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

help: error occurred in this function call
   ┌─ @local/cetz:0.2.0/src/process.typ:12:22
   │
12 │   (ctx, ..element,) = element-func(ctx)
   │                       ^^^^^^^^^^^^^^^^^

help: error occurred in this call of function `element`
   ┌─ @local/cetz:0.2.0/src/process.typ:58:12
   │
58 │     let r = element(ctx, el)
   │             ^^^^^^^^^^^^^^^^

help: error occurred in this call of function `many`
   ┌─ @local/cetz:0.2.0/src/canvas.typ:51:33
   │
51 │   let (ctx, bounds, drawables) = process.many(ctx, body)
   │                                  ^^^^^^^^^^^^^^^^^^^^^^^
```

</details>

johannes-wolf:
Fixed.

johannes-wolf:
I am not sure if the anchors of corners with different x/y radii are correct yet.

src/path-util.typ
```diff
@@ -123,37 +127,91 @@
...
 
-  let traveled-length = 0
-  for s in segments {
-    let part-length = segment-length(s)
+  let abs = type(t) != ratio
+  let target = calc.max(0, if type(t) == ratio {
+    t * length(segments) / 100%
+  } else {
+    t
+  })
```
fenjalien, line 185: We should check that `t` is in an acceptable range before this, instead of snapping to the start of the segment. There also doesn't seem to be a check for if the distance is greater than the length of the path.

src/path-util.typ
```diff
@@ -123,37 +127,91 @@
...
-    traveled-length += part-length
+    traveled += part
   }
+
+  return segment-end(segments.last())
 }
 
 /// Get position and direction on path
 ///
+/// TODO: Remove this function!
```
fenjalien, line 213: why? I might be using it for path anchors

src/path-util.typ
```diff
@@ -71,33 +68,36 @@
...
 /// -> vector: Position on the polyline
-#let point-on-polyline(s, t) = {
-  if t == 0 {
+#let _point-on-polyline(s, t) = {
+  if t == 0 or t == 0% {
     return s.at(1)
-  } else if t == 1 {
+  } else if t == 100% {
     return s.last()
   }
```
fenjalien, line 78: Check that `t` is between 0% and 100% or 0 and the length of the polyline.

src/path-util.typ
```diff
@@ -106,14 +106,18 @@
...
+/// - t (float,ratio): Absolute (float) or relative (ratio) position
+///
 /// -> vector: Position on segment
-#let point-on-segment(s, t) = {
-  let (type, ..pts) = s
-  if type == "line" {
-    return point-on-polyline(s, t)
-  } else if type == "cubic" {
-    let len = bezier.cubic-arclen(..pts) * calc.min(calc.max(0, t), 1)
+#let _point-on-segment(s, t) = {
```
fenjalien, line 112: Check that `t` is between 0% and 100% or 0 and the length of the segment.

src/path-util.typ
```diff
@@ -123,37 +127,91 @@
...
+  for s in segments {
+    let part = _segment-length(s)
+
+    // This segment contains target
+    if traveled <= target and target <= traveled + part {
+      return if abs {
+        _point-on-segment(s, target - traveled)
+      } else {
+        _point-on-segment(s, (target - traveled) / part * 100%)
+      }
```
fenjalien, line 202: Why not keep the distance as a float?

src/draw/shapes.typ
```diff
@@ -683,19 +685,37 @@
...
-            south-east: (b.at(0), a.at(1)),
+            south-east: se,
             south: (center.at(0), a.at(1)),
-            south-west: a,
+            south-west: sw,
             west: (a.at(0), center.at(1)),
-            north-west: (a.at(0), b.at(1)),
+            north-west: nw,
             center: center
           ).at(anchor)
```
fenjalien, line 720: Would it be easier to generate the compass anchors with border anchors? At least easier to understand whats going on...

src/draw/shapes.typ
```diff
@@ -670,6 +670,8 @@
     ctx => {
       let ctx = ctx
       let (ctx, a, b) = coordinate.resolve(ctx, a, b)
+      assert(a.at(2) == b.at(2),
+        message: "Both rect points must have the same z coordinate.")
```
fenjalien, line 674: "Both rectangle points must have the same z value."

src/draw/shapes.typ
```diff
@@ -705,16 +725,60 @@
...
+        segments += (path-util.line-segment((p1, p2)),)
+        segments += corner-arc(ne, 90deg, 0deg, p2)
+        segments += (path-util.line-segment((p3, p4)),)
+        segments += corner-arc(se, 0deg, -90deg, p4)
+        segments += (path-util.line-segment((p5, p6)),)
+        segments += corner-arc(sw, -90deg, -180deg, p6)
+        segments += (path-util.line-segment((p7, p0)),)
+
+        drawable.path(segments, fill: style.fill, stroke: style.stroke, close: true)
+      }
```
fenjalien, line 781: I don't understand how this works and it scares me... Could you tidy it up a bit and some more detailed comments for each step?

src/path-util.typ
johannes-wolf: The implementation is very bad. We can change point-on-path to give us the point + the direction vector which is faster and more precise (for curves).

src/path-util.typ
johannes-wolf: I want to support absolute and relative distances.

src/path-util.typ
fenjalien: Yea thats what I was thinking. At least make note of its (and others) behaviour on this. tbh I don't think we really need to prioritise performance this much at the moment.

fenjalien (commented):
Update the changelog as well

src/path-util.typ
```diff
@@ -123,37 +132,64 @@
...
 
 /// Get position on path
 ///
 /// - segments (array): List of path segments
-/// - t (float): Position (from 0 to 1)
-/// -> vector: Position on path
+/// - t (int,float,ratio): Absolute position on the path if given an
+///   float or integer, or relative position if given a ratio from 0% to 100%
+/// -> none,vector: Position on path as vector clamped to the paths begin/end
+///    position.
```
fenjalien, line 144: Clamping note is in the wrong place, it should be on `point-on-segment`. `point-on-path` panics if it is out of range.

fenjalien (commented):
changelog!

src/path-util.typ
johannes-wolf: Fixed. Forgot to add the CHANGELOG to the commit.

fenjalien (approved):

PR #356, opened 2023-11-27
Add shear-x and shear-y

Kreijstal:
We add both shears since they're not available.

### DIFF
--- a/src/matrix.typ
+++ b/src/matrix.typ
@@ -38,6 +38,22 @@
    (0, 0, 0, 1))
 }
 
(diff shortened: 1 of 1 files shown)


johannes-wolf:
Do you also want to add a transformation function in `/src/draw/transformtations.typ` for simpler use? Something like:
```typst
#let shear(x-factor, y-factor, z-factor) = {
  ..
}
```

johannes-wolf:
Closed because of no response from author.

PR #360, opened 2023-11-28
arc: Set previous point to arc end

johannes-wolf:
The `arc` command should update the previous point to its end-point.
This might be inconsistent, though, as there are no other commands setting the previous point to a non user-provided one. Should we add an option to enable/disable this @fenjalien?


fenjalien:
Yes this makes sense so then the previous coordinate can be used to form a path more easily. Lets have it on by default and give a style option to turn it off

src/draw/shapes.typ
```diff
@@ -215,6 +215,7 @@
 /// == Keys
 ///   #show-parameter-block("radius", ("number", "array"), [The radius of the arc. An elliptical arc can be created by passing a tuple of numbers where the first element is the x radius and the second element is the y radius.], default: 1)
 ///   #show-parameter-block("mode", ("string",), [The options are: "OPEN" no additional lines are drawn so just the arc is shown; "CLOSE" a line is drawn from the start to the end of the arc creating a circular segment; "PIE" lines are drawn from the start and end of the arc to the origin creating a circular sector.], default: "OPEN")
+///   #show-parameter-block("update-position", ("bool",), [Update the current canvas position to the arc's end point (anchor `"arc-end"`)], default: true)
```
fenjalien, line 218: Make a note that this overrides the default behaviour of setting it to the most recently coordinate given by the user and that it allows chaining of elements

PR #367, opened 2023-12-03
chore(imports): changed paths absolute -> relative

Andrew15-5:


johannes-wolf:
@Andrew15-5 this got auto-closed by my force push… so the close was by accident!
I am not pro merging this, though. What do you think, @fenjalien? Use case is having cetz in a sub-dir and importing it without @local or @preview installation.

Andrew15-5:
BTW, I re-did the paths on my branch (https://github.com/Andrew15-5/cetz/tree/relative-imports) from `0.2.0` branch instead of from `plot-annotations` (because `0.2.0` is the central branch that will be merged to `master`). If you want, I can open a new PR, and we can discuss it over there. I'm asking since there is no high change of being merged, so what would be the point in creating a second one that could be also unmerged (although this one was an accident).

Andrew15-5:
Yes, I also thought about this, because changes in my PR probably won't stop other branches from merging into `0.2.0` with no conflicts. So all other branches must be explicitly rebased/merged `0.2.0` into them to "not lose relative paths patch".

Andrew15-5:
Ok, let's discuss a little bit. So your only "anti-relative path" argument is that relative paths are ugly (which I agree with), right? My arguments are:
- local files/dirs autocompletion (in IDE/text editor) only works for relative paths (if I were to develop the cetz pacakge myself);
- with relative paths I can copy the whole cetz package folder and put it in my Typst project and import it as:

  ```typ
  #import "./cetz/src/lib.typ" as cetz
  ```

Andrew15-5:
As a compromise, you can make a script that will convert all absolute paths to relative ones. And this script will be used before creating a PR to publish a new version. So you copy necessary stuff from this repo to a new dir in `typst/packages` fork, then you run the script that converts all paths, open PR, it gets merged, done!

fenjalien:
I'm also against using relative paths over absolute. The script may also introduce bugs and unintended behaviour so we would have to run, then test it, then fix anything that broke. 

Why not create the script yourself and just run it on a local copy?

Andrew15-5:
> If you want to have some modified version that should only get pulled by a single project, you could give that version some special name and/or version number by modifying typst.toml before installation.

Imagine after a while having 100 different cetz `@local` packages, awful! Even having 2 isn't pretty (not *as all thing should be*). If I need to share the project or archive it I would hate going to `@local` and copying _that one version of a package_. And when recreating the project, I would have to split some stuff into 2 separate dirs to be able to compile the project again, you know?

johannes-wolf:
Ok. I am open to adding an option to the `package` script that sed's all includes and changes them to relative ones for all files copied. What do you think?

johannes-wolf:
@Andrew15-5 please test: https://github.com/johannes-wolf/cetz/pull/402

PR #369, opened 2023-12-04
path: Normalize paths

johannes-wolf:


src/path-util.typ
```diff
@@ -176,6 +176,41 @@
...
+  for s in segments {
+    if new == () {
+      new.push(s)
+    } else {
+      let head = new.last()
+      let (kind, ..pts) = s
+
+      if kind == "line" and head.at(0) == kind {
+        // Merge consecutive line segments
+        new.last() += pts.slice(1)
```
fenjalien, line 195: Why are `pts` being sliced? Doesn't that drop the first point of the line?

src/path-util.typ
```diff
@@ -176,6 +176,41 @@
...
+      } else if segment-start(s) != segment-end(head) {
+        // Push a new line or line point if the current segment
+        // does not start where the previous segment ended
+        if head.at(0) == "line" {
+          new.last().push(segment-start(s))
```
fenjalien, line 200: Why not use `pts.first()` instead of `segment-start`?

fenjalien (changes requested):

fenjalien:
Oh and how could we go about adding a test for this? Or would it be covered by a merge path test?

johannes-wolf:
> Oh and how could we go about adding a test for this? Or would it be covered by a merge path test?

Fixed. I've added a test case to the `merge-path` test.

fenjalien (approved):

Issue #234, opened 2023-09-30
Drawing well-centered lines between elements


TimotheAlbouy:
Currently, there is no simple way to draw a "well-centered line" between 2 elements in CeTZ.

I call a line between two elements "well-centered" if it follows the same path as the infinite line that passes through the centers of both elements, but it just touches the edges of the 2 elements instead. I know that this may not be very clear, so this is the kind of things that I want to achieve (this was produced in TikZ):

<img src="https://github.com/johannes-wolf/cetz/assets/38692015/4f74fe12-e407-4ef4-9b23-f65c4fa2b8a7" width="300">

The solution should also be element- and position-agnostic, so no cheating like for instance assuming that the elements are circles of a given radius, or assuming their position and inferring a given line angle.

Doing this in TikZ is very easy:
```latex
\node[draw, circle] (A) at (0,0) {arbitrary text};
\node[draw, align=left] (B) at (2,3) {arbitrary\\text};
\draw[<->] (A) -- (B);
```

So the solution in CeTZ should be as intuitive as the TikZ solution. Something like:
```typst
content((0,0), [arbitrary text], name: "A", frame: "circle")
content((2,3), [arbitrary\ text], name: "B", frame: "rect")
line("A", "B", mark: (end: ">", start: ">"))
```

Such a feature could be very helpful for a variety of things, such as drawing pretty box-and-arrow diagrams. It could also be useful for the tree CeTZ library (see [this other issue](https://github.com/johannes-wolf/cetz/issues/225)) to replicate the behaviour of the tree TikZ library (the lines between the elements are "well-centered", and not drawn between the `top` and `bottom` anchors of the elements).

See [this Discord thread](https://discordapp.com/channels/1054443721975922748/1154880386942390366) for more info.

TimotheAlbouy:
So now, let's talk about how this feature could be implemented.

Currently, `line("A", "B", mark: (end: ">", start: ">"))` resolves the `"A"` and `"B"` parameters of the `line()` function as the `"A.default"` and `"B.default"` anchors, which respectively correspond to the centers of the circle and the rectangle. So this code would produce a line that goes from the center of the circle to the center of the rectangle, which is not what we want. This current behaviour of CeTZ could be redefined to implement the feature of this post.

[...]

PR #376, opened 2023-12-04
hide: Hide elements

johannes-wolf:
This adds a function for hiding elements.
Hidden elements act like normal elements, despite not getting rendered and not producing a bounding box.

This can be handy for calculating intersections between elements that should not get drawn, which is currently only possible by using the internal API.

@fenjalien, I think we should add some Typst native functions back as aliases in draw.typ:
```
#let typst-line = line
#let typst-hide = hide
...
```

Makes #234 at least possible. A `line(..)` API will follow in a separate PR.


src/draw/grouping.typ
```diff
@@ -13,6 +13,41 @@
 
 #import "transformations.typ": move-to
 
+/// Hide an element.
+///
+/// Hidden elements act like normal elements, despite not getting
+/// rendered and not producing a bounding box.
```
fenjalien, line 19: "Hide*s*..."

"Hidden elements are not drawn to the canvas, are ignored when calculating bounding boxes and discarded by `merge-path`. All other behaviours remain the same as a non-hidden element."

fenjalien (approved):
Add a comment about hidden paths to merge-path's docs as well!

PR #377, opened 2023-12-05
Rework style (again)

fenjalien:
- Reworks styling to work in a kind of simpler way.
- Fixes wrong styling in an example in the manual
- Added detailed description of how styling works


src/draw/shapes.typ
```diff
@@ -665,17 +666,21 @@
       (calc.max(from.at(0), to.at(0)), calc.max(from.at(1), to.at(1)))
     )
 
-    let style = styles.resolve(ctx.style, style)
-    if help-lines {
+    let style = styles.resolve(ctx.style, merge: style, root: "grid", base: (
```
johannes-wolf, line 669: Any reason not having this in the default style dictionary?

src/lib/axes.typ
```diff
@@ -30,6 +30,18 @@
     stroke: (paint: gray, dash: "dotted"),
     fill: none
   ),
+  x: (
```
johannes-wolf, line 33: If we add `x` and `y`, we should also add `x2` and `y2`.

src/styles.typ
```diff
@@ -108,68 +124,96 @@
...
-            v = parent.at(k)
-            if v != auto {
-              dict.insert(k, v)
-              break
+      let is-dict = type(v) == dictionary
+      if is-dict or v == auto {
+        for ancestor in ancestors {
+          if k in ancestor {
+            // If v is auto and the ancestor's value is not auto, update v.
+            if (ancestor.at(k) != auto and v == auto) {
```
johannes-wolf, line 181: Nitpick: remove parentheses.

johannes-wolf (changes requested):
Some small things only.
The changed `arc` test image does not match the output. Current reference images are still Typst 0.9.0.

src/draw/shapes.typ
fenjalien: It doesn't really matter where the default of an element's style is. Really `styles.default` can just have stroke, fill, radius, padding, shorten and mark keys in it. Then use `base` for each element and it'll work just the same...

tests/axes/test.typ
```diff
@@ -1,14 +1,19 @@
...
-  set-style(axes: (padding: .75))
-  set-style(axes: (x: (stroke: red)))
-  set-style(axes: (y: (stroke: green, tick: (stroke: blue, length: .3))))
+  set-style(axes: (
+    stroke: blue,
+    padding: .75,
+    x: (stroke: red),
+    y: (stroke: green, tick: (stroke: blue, length: .3))
+  ))
+  // set-style(axes: (padding: .75))
```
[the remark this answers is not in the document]
fenjalien: removed

src/styles.typ
```diff
@@ -26,78 +26,94 @@
...
+  circle: (
+    radius: auto,
+    stroke: auto,
+    fill: auto
+  ),
+  rect: (
+    stroke: auto,
+    fill: auto,
   ),
-  mark: _default-mark,
```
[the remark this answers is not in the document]
fenjalien: Moved comments and removed

src/lib/axes.typ
fenjalien: There is no x2 and y2 but top, right, bottom and left. Added the defaults within the first axis-settings loop.

src/styles.typ
fenjalien: Removed

PR #378, opened 2023-12-06
line: Draw line between element borders

### LINKED ISSUE #234: Drawing well-centered lines between elements

johannes-wolf:
Fixes #234

If last and or first argument of `line` is an element name, use that for intersection test.
Saves complete elements in `ctx.nodes`.


src/anchor.typ
```diff
@@ -109,12 +109,23 @@
       continue
     }
     pts += intersection.line-path(..test-line, drawable)
+  }
+  assert(pts.len() > 0, message: strfmt("{} {} {}", test-line, drawables, angle))
```
fenjalien, line 113: Remove this

src/draw/shapes.typ
```diff
@@ -582,9 +594,72 @@
...
+        vector.as-vec(pt, init: (0,0,0,1))).slice(0, 3)
+    }
+  }
   
   return (ctx => {
+    let first-elem = pts.first()
+    let first-is-elem = coordinate.resolve-system(first-elem) == "element"
+
+    let last-elem = pts.last()
+    let last-is-elem = coordinate.resolve-system(last-elem) == "element"
```
fenjalien, line 635: Use the results of `resolve-system` from line 596 above.

fenjalien (changes requested):

tests/intersection/test.typ
```diff
@@ -69,7 +69,7 @@
     content((0, 0), [Das ist\ ein Text!], frame: "circle", name: "a")
     content((2, 1), [Hallo!], frame: "circle", name: "b")
     // Invisible intersection line
-    line("a", "b", stroke: none)
+    line("a.default", "b.default", stroke: none)
```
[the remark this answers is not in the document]
johannes-wolf: Because we do not want to calculate intersections here.

src/draw/shapes.typ
johannes-wolf: What do you mean? I am using the result to compare against "element"

src/draw/shapes.typ
```diff
@@ -556,8 +556,20 @@
 /// line((-1.5, 0), (1.5, 0))
 /// line((0, -1.5), (0, 1.5))
 /// line((-1, -1), (-0.5, 0.5), (0.5, 0.5), (1, -1), close: true)
-/// ```) 
-/// 
+/// ```)
+///
+/// If the first or last coordinate given is the name of an element,
+/// the intersection of that element and a line from the first or last
+/// two coordinates given is used as coordinate. This is useful to
```
fenjalien, line 563: What about "If the first or last points are given as the name of an element, that has a "default" anchor, the intersection of that element's border and the line."

fenjalien (commented):
I'm having double thoughts about not using border anchors for this. If we do custom elements can define their own border instead relying on something you can't easily configure...

johannes-wolf:
> I'm having double thoughts about not using border anchors for this. If we do custom elements can define their own border instead relying on something you can't easily configure...

I do not get what you mean. Custom elements should be able to override the behavior of how the intersection is determined?

PR #379, opened 2023-12-06
intersections: Support named elements

### LINKED ISSUE #230: Compute the intersections of elements using their names
Currently, to compute the intersections of some elements, we pass the elements directly to the body argument of the `intersections()` function. However, this has several disadvantages.

Firstly, it is not possible to compute the intersections of elements that were already displayed before calling `intersections()` without displaying the elements a second time, which produces uglier edges on the elements. See [this Discord thread](https://discordapp.com/channels/1054443721975922748/1157333272959914004) for more info.

Secondly, it means that if we do a function that computes the intersections of several elements, then we HAVE to pass these elements directly, and not just their names. This, I think, goes against the design principles of the library, as the name is the idiomatic way of referring to an element in CeTZ (e.g., we pass names to the `line()` function, not elements directly). Moreover, we have no way of retrieving an element using just its name in CeTZ (even using the context). See [this Discord thread](https://discordapp.com/channels/1054443721975922748/1157346054619017336) for more info.

So for me, the solution would be to change how the `intersections()` function works, and pass as parameters just names and not elements.

johannes-wolf:
The `intersections(..)` function now accepts names of elements (in addition with elements)
to calculate intersections between.

Fixes #230.


src/draw/grouping.typ
```diff
@@ -106,9 +138,10 @@
 /// The default anchor is "center" but this can be overridden by using `anchor` to place a new anchor called "default".
 ///
 /// - body (elements, function): Elements to group together. A least one is required. A function that accepts `ctx` and returns elements is also accepted.
-/// - name (none, string):
-/// - anchor (none, string):
-/// - ..style (style):
+/// - name (none, string): Name of the group element
+/// - anchor (none, string): Anchor to position the group and it's children relative to. For translation the difference between the groups `"center"` anchor
+///   and the passed anchor is used.
+/// - ..style (style): Style key-value pairs
```
fenjalien, line 144: I haven't been documenting these arguments as they always act the same on elements and are documented elsewhere. They still show up on the function definition even if they are not written here. 

Maybe the anchor description is okay as its behaviour should be noted but it could be moved to under the anchor heading.

johannes-wolf: Reverted. I left the anchor description, though.

PR #380, opened 2023-12-08
manual: Some small fixes and improvements

johannes-wolf:


fenjalien:
Why should `ctx` be opaque? And can we just hold documentation fixes until we're done adding feature docs as I'm sure they'll be other mistakes to correct (they can be noted in this PR).

johannes-wolf:
… as in opaque to the user, because it is undocumented and unstable.

PR #389, opened 2023-12-10
Nested Element Access

johannes-wolf:
This adds nested element access using “.” syntax.
Nested elements are added to a group's private `nodes` dictionary first, from which the
group copies them to the new context object.

I implemented paths as “flat” strings, as real element nesting might not be needed.

The new group-private `nodes` dict is needed to know which elements got added or overridden insides the group.


johannes-wolf:
Closing in favor of #523.

PR #402, opened 2023-12-13
scripts: Add script for changing all abs paths

johannes-wolf:
This allows installing cetz with all paths in all *.typ files changed to relative ones.
@Andrew15-5: Please test this via `just package <your-directory> --relative-paths`.

As soon as Typst supports changing the package path, this cursed script must get removed.

### DIFF
--- a/justfile
+++ b/justfile
@@ -4,8 +4,8 @@
 gallery_dir := "./gallery"
 test_dir := "./tests"
 
-package target:
-  ./scripts/package "{{target}}"
+package target *options:
+  ./scripts/package "{{target}}" {{options}}
 
 install target="@local":
   ./scripts/package "{{target}}"
--- a/scripts/package
+++ b/scripts/package
@@ -24,7 +24,7 @@ else
 fi
 
 if (( $# < 1 )) || [[ "${1:-}" == "help" ]]; then
-  echo "package TARGET"
+  echo "package TARGET [--relative-paths]"
   echo ""
   echo "Packages all relevant files into a directory named '${PKG_PREFIX}/<version>'"
   echo "at TARGET. If TARGET is set to @local, the local Typst package directory"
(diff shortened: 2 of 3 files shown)


Andrew15-5:
BTW, it's not great when you import the same published package multiple times. I had an issue (with cetz) where I thought I've changed the version, but actually I imported the same old version in different files. So I advise you to import external libraries somewhere once and then import that local file everywhere where you need external library API.

You can very easily forget to update all files that import the same package.

```
src/anchor.typ
1:#import "@preview/oxifmt:0.2.0": strfmt

src/draw/grouping.typ
1:#import "@preview/oxifmt:0.2.0": strfmt

src/coordinate.typ
3:#import "@preview/oxifmt:0.2.0": strfmt

src/util.typ
1:#import "@preview/oxifmt:0.2.0": strfmt
```

Andrew15-5:
I test some basic stuff, and it works:

```typ
#import "./cetz/0.2.0/src/lib.typ" as cetz

#cetz.canvas({
  import cetz.draw: *
  circle(())
})

#cetz.canvas({
  import cetz.draw: *
  import cetz.plot
  plot.plot(size: (5, 5), {
    plot.add(domain: (-10, 10), x => x * x)
  })
})
```

https://github.com/johannes-wolf/cetz/pull/367#issuecomment-1853626270:

> I think Typst checks/imports everything first, before compiling, so the test is just "does it compile a document?" Although I'm not sure.

Indeed, it [checks for everything](https://discord.com/channels/1054443721975922748/1088371867913572452/1184637134984970240), therefore if there is an invalid import, then the error must show up. So any "import cetz and compile" test should prevent any relative path bugs.

Andrew15-5:
[Note this](https://discord.com/channels/1054443721975922748/1088371867913572452/1184638412398002267):

> So the only thing that won't be noticed right away [by the compiler] is if `if` statement block isn't run or loop isn't run or a function isn't called, but all root imports/includes will 100% be checked.

johannes-wolf:
> [Note this](https://discord.com/channels/1054443721975922748/1088371867913572452/1184638412398002267):
> 
> > So the only thing that won't be noticed right away [by the compiler] is if `if` statement block isn't run or loop isn't run or a function isn't called, but all root imports/includes will 100% be checked.

So what? Test cases should run all code (this is a guess).

Andrew15-5:
> So what? Test cases run all code (I guess). Imports in functions or `if`s are fine.

Then we are golden (good to go). If you run your regular tests after applying relative paths and before publishing the package, then no additional tests are necessary.

PR #408, opened 2023-12-17
Removed leading whitespaces

Andrew15-5:
```sh
rg -l0 '[ \t]+$' | xargs -0 sd '[ \t]+$' ''
```

PR #409, opened 2023-12-17
Made documentations 2D alignment consistent

Andrew15-5:
Alignment terms aren't used with "-" in normal text, only in programs like Typst. And the overall use of alignment terms was inconsistent.

### DIFF
--- a/src/draw/transformations.typ
+++ b/src/draw/transformations.typ
@@ -216,7 +216,7 @@
 /// circle((5,5))
 /// ```)
 ///
-/// - from (coordinate): Bottom-Left corner coordinate
+/// - from (coordinate): Bottom left corner coordinate
 /// - to (coordinate): Top right corner coordinate
 /// - bounds (vector): Viewport bounds vector that describes the inner width,
 ///   height and depth of the viewport


johannes-wolf:
Can you give a reference? I think both versions are correct (@fenjalien?). The second change is def. incorrect, as it refers to the anchor names, which are hyphenated.

src/draw.typ
```diff
@@ -492,13 +492,13 @@
...
-/// - bottom: Mid between bottom-left and bottom-right
+/// - top: Mid between top left and top right
+/// - left: Mid between top left and bottom left
+/// - right: Mid between top right and bottom right
+/// - bottom: Mid between bottom left and bottom right
 ///
-/// - a (coordinate): Bottom-Left coordinate
-/// - b (coordinate): Top-Right coordinate
+/// - a (coordinate): Bottom left coordinate
+/// - b (coordinate): Top right coordinate
```
Andrew15-5, line 501: > The second change is def. incorrect, as it refers to the anchor names, which are hyphenated.

You mean all this? I've noticed that in all other places, dashed words are enclosed in "\``". So if some of these words are referring to the Typst value/argument, then they also must be enclosed with "``", again, to make documentation more consistent.

Andrew15-5:
> Can you give a reference? I think both versions are correct.

https://www.collinsdictionary.com/dictionary/english/bottom-right
https://thecontentauthority.com/blog/how-to-use-bottom-right-in-a-sentence

Although, it's kinda hard to find something concrete. And people indeed do use dash sometimes, and no one is correcting them (but even if this was an error no one probably would correct them as this wasn't the main problem). But the main point is that this dash is inconsistent. You should choose one style. I'm down with either one.

fenjalien:
Whatever is correct, it shouldn't be going in to master, it should be going in to the 0.2 branch.

johannes-wolf:
I suggest keeping this (+ the space PR) open, until all the PRs for 0.2.0 got merged.

Andrew15-5:
> Whatever is correct, it shouldn't be going in to master, it should be going in to the 0.2 branch.

I already addressed that in https://github.com/johannes-wolf/cetz/pull/409#issuecomment-1860953244:
> (I then also change the base branch.)

Andrew15-5:
> I suggest keeping this (+ the space PR) open, until all the PRs for 0.2.0 got merged.

In other words, you feel like future commits to 0.2.0 will negate changes from these 2 PRs? If so, then I agree.

@johannes-wolf, does this mean that I should change back the base branch in #408 (and wait)?

johannes-wolf:
> > I suggest keeping this (+ the space PR) open, until all the PRs for 0.2.0 got merged.
> 
> In other words, you feel like future commits to 0.2.0 will negate changes from these 2 PRs? If so, then I agree.
> 
> @johannes-wolf, does this mean that I should change back the base branch in #408 (and wait)?

No, we just do not merge this _yet_.

johannes-wolf:
@Andrew15-5 can you rebase on `doc-adjust-ctx`? Then we can merge this PR into that branch.

Andrew15-5:
I have run the `\rg -i '(bottom|top|left|right)[- ](bottom|top|left|right)` to check for not-in-code `-`-words and I have found only one inconsistency.

PR #385, opened 2023-12-09
Mark improvements

fenjalien:
This does another rework of marks, specifically:

- Move marks into a dictionary of functions.
- Marks are initially drawn at (0,0) pointing towards the left then translated and rotated in the correct direction.
- Rework path utils for finding points on segments
- Unifies mark placement on paths with a single function `place-marks-along-path`

Current progress:
- [x] Proof of concept
- [x] Tip offset
- [x] Improve cubic segment handling (flexing, shortening)
- [x] Add to all mark enabled elements
- [ ] Add more marks
- [ ] Add more styling

I'm basing a lot of the maths and code on #333.

PR #416, opened 2023-12-20
Mark Improvements

johannes-wolf:
Rebased and improved copy of #385.

**TODOs:**
- [x] Setting `flex: false` is broken (see tests, wrong mark offset)
- [x] Add open triangle mark
- [x] Add curved triangle mark (barbed)
- [x] Add +, x and * marks
- [x] Allow marks being longer then their path; do not panic if a path is very short
- [x] Fix reversed mark offsetting
- [x] Fix z-axis marks
- [x] Add back “curve snapping” (cubic shorten curve correction)
- [x] Add offset
- [x] Add position (override a mark's position) - Allows relative or absolute mark placement on a path

@fenjalien I've pushed your branch and added some fixes, improvements and missing marks.

I've added linear path extrapolation, which is what TikZ seems to do, too. Marks can “grow” out of the path now.


johannes-wolf:
I ported 75972d7ddade928dac1742ae31dc3fdd1ecacf17 over.

PR #447, opened 2024-01-10
Use ratios for lerp coordinates

### LINKED ISSUE #442: Change linear interpolation coordinates to use ratio instead of float
Ratio: Relative distance interpolation, Number: Absolute distance interpolation

johannes-wolf:
Fixes #442.

This change should make the intent of coordinates like `(a, 50%, b)` more clear.
In a future change, I would like to default `abs` on the type (ratio or number) used.


johannes-wolf:
@fenjalien I guess it would be nice to add this to 0.2.0 too, to start supporting the proposed ratio syntax, so a future breaking change (hopefully) won't affect too many documents.

fenjalien:
Why not remove `abs` now? Ratio can be the new relative distance and a number is absolute?

johannes-wolf:
Rebased on the decorations PR. This introduces a breaking change, making all interpolation with numbers (float, length) absolute, while requiring rations for the old behaviour.

tests/coordinate/lerp/test.typ
fenjalien, line 1: Add a test case for the `angle` parameter

johannes-wolf: Done.

PR #468, opened 2024-01-17
hobby: Fix closed two point hobby curve

### LINKED ISSUE #462: Hobby should only return a straight line for two points if close: false
Because of the following code
```
if points.len() == 2 {
    let (a, b) = points
    return ((a, b, a, b),)
  }
```
hobby will always draw a straight line with two points given. But this should only happen if `close: false`.

johannes-wolf:
Fixes #462.


fenjalien:
@Enivex do the tests look right?

Enivex:
> @Enivex do the tests look right?

I would say so, yes

fenjalien (approved):

PR #479, opened 2024-01-21
Rework Plot Axes

johannes-wolf:
Some cleanup of the axis drawing code. More refactoring should follow, but in a follow-up PR.

Axis styling documentation is not good yet!
New features:
- Per axis coloring for scientific plots
- Axis padding for both scientific and school-book plots
- Axis "overshoot" for school-book plots
- Customizable axis, tick and grid z-layer
- One can disable the shared "0" tick

I am not happy with the shared "0" tick drawing, this should get improved!


src/process.typ
```diff
@@ -56,14 +56,16 @@
   let drawables = ()
   let bounds = none
 
-  for el in body {
-    let r = element(ctx, el)
-    if r != none {
-      if r.bounds != none {
-        bounds = aabb.aabb(r.bounds, init: bounds)
+  if body != none {
```
fenjalien, line 59: Why not panic? Why pass silently?

tests/chart/ref.png
fenjalien, line 1: Is the padding on the second chart okay?

fenjalien (changes requested):
Are you happy with the state of `axes.typ`? I haven't focused on it as I don't mind but I can take a closer look/be more picky if you want.

Are you going to fix the docs in this PR?

UPdate the changelog!

src/lib/axes.typ
```diff
@@ -3,29 +3,66 @@
...
 #import "/src/vector.typ"
 #import "/src/styles.typ"
+#import "/src/process.typ"
+#import "/src/drawable.typ"
+#import "/src/path-util.typ"
 
 #let typst-content = content
 
-// Global defaults
+/// Default axis style
```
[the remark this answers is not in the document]
johannes-wolf: Yes.

src/process.typ
johannes-wolf: An empty (group) body should be legal. Example:
```typst
group({
  if ... {
    draw(...)
  }
})
```
Has the body set to `none` if the if-condition is false. This should not panic, but still create a group (or any other grouping element).

johannes-wolf:
> he docs in this PR?

No, I am not yet happy but would like to do further changes later.
Docs for styling are now added (although they are quite minimal).

src/process.typ
fenjalien: I would prefer the check to be in group then instead of `process.many` which should always expect an array

johannes-wolf: But isn't `none` or `()` like kind of the same? I mean, the ctx function could also return `none` which would be a second check… I favour having this at the lower level, I would even propose allowing passing functions to `process.element` and `*.many`.

johannes-wolf: Should be solved with #491.

fenjalien (approved):
Make sure to rebase this first, as `resolve-body` is still in `grouping.typ`

PR #491, opened 2024-02-10
grouping: Handle none as empty array

johannes-wolf:
Allows passing `none` or `ctx => {}` to grouping functions (`group`, `on-layer`, ...).
This is very useful when having conditional draw commands insides a group.

Example:
```
group(ctx => {
  if .. {
    // If this does not get evaluated, cetz currently fails because the group returned `none`.
    // Current workaround is writing () at the end of such groups to auto-join an empty array.
  }
})
```

This PR is an improved implementation of the one rejected from #479.


fenjalien:
How do you want to review/merge this? Comments here then merge this before #479?

johannes-wolf:
> How do you want to review/merge this? Comments here then merge this before #479?

Yes, independent of #479.

src/draw/grouping.typ
```diff
@@ -13,6 +13,29 @@
...
+  } else {
+    body
+  }
+
+  return if body == none {
+    ()
+  } else {
+    assert(type(body) == array)
+    body
+  }
```
fenjalien, line 36: I would prefer styling like:
```
if type(body) == function {
  body = body(ctx)
}
if body == none {
  body = ()
} // else { // could be idk
assert(type(body) == array, message: "body must be of type array, none, or a function that returns an array or none, got: " + repr(body))
return body
```

src/draw/grouping.typ
```diff
@@ -109,7 +132,7 @@
           message: "No such element '" + elem + "' in elements " + repr(ctx.nodes.keys()))
         named-drawables.push(ctx.nodes.at(elem).drawables)
       } else {
-        for sub in elem {
+        for sub in resolve-body(ctx, elem) {
```
fenjalien, line 135: I'm not sure this should work as in this situation:
```
intersections("i", {
  line(...)
  line(...)
})
```
`elem` would be the element function which should be passed directly to `process.element`

src/draw/grouping.typ
```diff
@@ -432,13 +453,14 @@
...
 ///
 /// - layer (float, integer): The layer to place the elements on. Elements placed without `on-layer` are always placed on layer 0.
-/// - body (elements): Elements to draw on the layer specified.
+/// - body (elements, function): Elements to draw on the layer specified. A function that accepts `ctx` and returns elements is also accepted.
 #let on-layer(layer, body) = {
-  assert(type(layer) in (int, float), message: "Layer must be a float or integer, 0 being the default layer. Got: " + repr(layer))
-  assert(type(body) in (function, array))
-  return (ctx => {
-    let (ctx, drawables, ..) = process.many(ctx, if type(body) == function { body(ctx) } else { body })
+  if body == none { body = () }
```
fenjalien, line 458: This isn't needed

src/draw/grouping.typ
fenjalien: Besides theres an assert requiring at least one element on line 93/116

src/draw/grouping.typ
fenjalien: Also this should probably be moved to the util file

src/draw/grouping.typ
```diff
@@ -192,21 +215,19 @@
...
   (ctx => {
     let style = styles.resolve(ctx.style, merge: style.named(), root: "group")
 
     let bounds = none
     let drawables = ()
     let group-ctx = ctx
     group-ctx.groups.push((anchors: (:)))
-    (ctx: group-ctx, drawables, bounds) = process.many(group-ctx, if type(body) == function {body(ctx)} else {body})
+
+    (ctx: group-ctx, drawables, bounds) = process.many(group-ctx, resolve-body(ctx, body))
```
fenjalien, line 230: Should we be passing `group-ctx` instead of `ctx`?

src/draw/grouping.typ
```diff
@@ -432,13 +453,14 @@
 /// ```)
 ///
 /// - layer (float, integer): The layer to place the elements on. Elements placed without `on-layer` are always placed on layer 0.
-/// - body (elements): Elements to draw on the layer specified.
+/// - body (elements, function): Elements to draw on the layer specified. A function that accepts `ctx` and returns elements is also accepted.
 #let on-layer(layer, body) = {
-  assert(type(layer) in (int, float), message: "Layer must be a float or integer, 0 being the default layer. Got: " + repr(layer))
-  assert(type(body) in (function, array))
```
fenjalien, line 438: Keep the layer assert here so its easier for the user to find. It might be worth checking the body anyway for the same reason? Then if all elements check that before calling the resolver, the resolver only has to check the output of the function...

fenjalien (changes requested):

src/draw/grouping.typ
johannes-wolf: The assert is still there.

src/draw/grouping.typ
johannes-wolf: I've removed the function support for on-layer again. If a user want's a context they can call `get-ctx`.

src/draw/grouping.typ
johannes-wolf: Checking the body for what?

fenjalien: So it is, bad github diff. I mean check that body is a function, array or none outside of the element function
```
assert(type(layer) ...)
assert(type(body) ...)
return ctx => ...
```
Then the error message will come from the user code and not hidden in all the function calls.

johannes-wolf: I've added `util.assert-body(...)` for early checking of the body argument(s).

fenjalien (approved):

Issue #251, opened 2023-10-13
[Feature Request] Make CetZ compatible with Polylux


cristobaltapia:
Hi, thanks for staring this project and give Typst some nice graphic library!

I was thinking of using CetZ in a presentation with [Polylux](https://github.com/andreasKroepelin/polylux) and in general it works very good. However, one feature that is not working is the use of the special functions of polylux to uncover elements (e.g. `only`, `uncover`), equivalent to "overlays" in LaTex+Beamer. It would be nice if CetZ could understand/use those functions within a figure, in order to make nice animations. Certainly this has not a great urgency, but it would be a nice addition :) .

johannes-wolf:
That is possible. It would be easier if Polylux exposed some public functions that take the location from `locate` as argument and return instead of calling locate itself. I will take a look into this.

cristobaltapia:
Thanks for considering this! Maybe I can open an issue in polylux about this.

andreasKroepelin:
(Polylux author here)
There is the fundamental issue that you can't `hide` lines or paths in Typst. That's why Polylux' `uncover`, `pause` etc. don't play nicely with CeTZ at the moment.

johannes-wolf:
Yes, that is because cetz lines or paths are not Typst content but dictionaries. But Polylux' `hide` returns content that must be visible in the document (`locate(..)`).

andreasKroepelin:
Well, you still can't hide them, regardless of how CeTZ represents them intermediately. But using `none` for strokes is a clever trick!

And you're right, Polylux should export some more overlay functions. Currently I'm thinking about good ways to provide information about the current subslide via a function. You could then do something like

```typ
#name-tba(args => {
  cetz.canvas({
    if "we're at the correct subslide" {
      cetz.stroke(red)
    } else {
      cetz.stroke(none)
    }
    // ...
  })
})
```

johannes-wolf:
What would real hiding them look like? I mean, I could also suppress the draw call, without the stroke/fill none hack, but there is no such flag as of now.

andreasKroepelin:
I was referring to the builtin `hide` function. So

```typ
#hide(line())
```

still shows a line, and this effecr propagates to CetZ.

laurmaedje:
That's pretty clearly a bug in Typst and not something polylux or cetz should have to work around.

Enivex:
> That's pretty clearly a bug in Typst and not something polylux or cetz should have to work around.

[...]

PR #492, opened 2024-02-11
Polylux Integration

johannes-wolf:
This PR adds integration for the Polylux module by @andreasKroepelin.
The code currently uses private functions from Polylux `logic.typ`, maybe those could be made public @andreasKroepelin (see `/src/lib/polylux.typ`)?

To use Polylux with CeTZ, the user has to pass the Polylux module to the canvas. This allows using different Polylux versions with CeTZ, without CeTZ having a dependency on Polylux.

For an example, see the test case.
The following Polylux functions are available from `cetz.polylux`:
- `only`
- `uncover`

@fenjalien, to be able to modify state (counters) I've added a list of functions to `ctx` that get executed in the `canvas` function and allow returning content prior to the canvas block. Since the content returned by counter updates is not visible, this should be fine. I could not come up with a better solution.

We could also implement this binding on the Polylux side.

With #494 there is a proof-of-concept for a plug-in API either Polylux or a package like `cetz-polylux` could use for providing the Polylux bindings.


andreasKroepelin:
Wow, that's cool! What Polylux functions specifically should be made public?

johannes-wolf:
> Wow, that's cool! What Polylux functions specifically should be made public?

See https://github.com/johannes-wolf/cetz/pull/492/files#diff-3d4562bd25428e8bd4aabe76e4f474ce42c427243eab52f8e09b121f1f687d4e

I had to replicate `_conditional-display` (which I guess is fine) and I am calling `_last-required-subslide` and `_check-visible` + using the `repetitions` and `subslides` counters.

OrangeX4:
As a lower-level foundational graphics package, should cetz rely on higher-level slide packages like polylux and touying, especially when adding a polylux parameter to cetz's canvas? I remain skeptical. I feel that even if support for slides is to be added, it should be done by finding a way for cetz to extract a better interface. For example, adding a parameter "current subslide index" to canvas and letting polylux depend on cetz, rather than the other way around.

For example, touying now supports the use of `only` in cetz (thanks for the code from polylux). So I thought there might be a better way to deal with the direction of dependency.

```typ
#import "@preview/touying:0.2.0": *
#import "@preview/cetz:0.2.0"

#(s.page-args.width = auto)
#(s.page-args.height = auto)
#(s.page-args.footer = none)
#let (init, slide) = utils.methods(s)
#show: init

#slide(repeat: 3, self => [
  #let (only,) = utils.methods(self)

  Current index: #self.subslide

  #cetz.canvas({
    import cetz.draw: *

    rect((-1,-1), (1,1), fill: red)
    only(2, circle((.5,.3), fill: blue))
    only("2-3", {
      line((.2,.1), (rel: (1,1.3)), (rel: (.7, -1.8)), fill: green, close: true)
      content((0, -2), [Hello])
    })
  })
])
```

image

OrangeX4:
Like here: https://github.com/johannes-wolf/cetz/issues/251#issuecomment-1773946454

fenjalien:
Closing this as @johannes-wolf and I figured last night that CeTZ should not be the package containing the integrations/plugins from other packages. They should be in the integrating package or a separate package. We are working on making integrations easier in #494, @OrangeX4 and @andreasKroepelin please join us there to discuss.

PR #493, opened 2024-02-11
mark: Support placing marks post-transform

johannes-wolf:
This allows placing marks _after_ the path got transformed.
Use-Case: drawing marks on scaled paths, drawing marks insides plots (which use a scaled viewport). The behavior is disabled by default and can be enabled by setting marks `post-transform` style key to `true`.

Maybe the implementation has too much copied code, but doing the transform insides `mark.typ` is also not that clean I think.


johannes-wolf:
… or should we even do this by default (without an option)? This is what TikZ does. Marks don't get scaled with transforms (at least that's what I've observed via testing it). What do you think @fenjalien?

johannes-wolf:
Together with #505 this would bring marks closer to how Tikz handles them.

src/draw/shapes.typ
```diff
@@ -312,20 +312,27 @@
...
+
     return (
       ctx: ctx,
       name: name,
       anchors: anchors,
-      drawables: drawable.apply-transform(
-        transform,
-        drawables,
-      )
+      drawables: drawables,
```
fenjalien, line 335: I think this implementation could be improved. Why not transform the drawables inside `place-marks`, then have transform the drawable if no marks are placed.

Although with this change I'm not sure about how the future of modifiers could be implemented for marks...

src/styles.typ
```diff
@@ -45,6 +45,9 @@
     /// set the default to `0` and to `auto` for the mark you want to
     /// shorten the path to. Set to `none` to disable path shortening.
     shorten-to: auto,
+    /// Place marks post path transformation. This is not honored per mark, but
+    /// for all marks on a path.
```
fenjalien, line 49: Whats stopping it from being a per-mark option?

fenjalien (changes requested):

src/styles.typ
johannes-wolf: Marks on transformed paths may be bigger, path shortening gets tricky because we would need to un-transform the path first if it was transformed by a previous mark… makes things necessary complicated IMHO.

From a user perspective, it makes no sense supporting it per mark.

src/draw/shapes.typ
johannes-wolf: As mentioned in the PR text, I do not think this is optimal. If the marks do the transform it is a bit too "hidden" and it won't save that many lines either. I'd rather have it done before calling `place-marks-along-path`.

manual.typ
```diff
@@ -278,6 +278,7 @@ line(..c)
 
 #doc-style.show-parameter-block("shorten-to", ("integer", "auto", "none"), [Which mark to shorten the path to when multiple marks are given. `auto` will shorten to the last mark, `none` will shorten to the first mark (effectively disabling path shortening). An integer can be given to select the mark's index.])
 
+#doc-style.show-parameter-block("post-transform", "bool", [If true, place marks post path transformation, which results in marks not affected by the current transformation.], default: false)
```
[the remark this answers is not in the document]
johannes-wolf: Maybe `transform-with-shape`? Or `apply-shape-transform`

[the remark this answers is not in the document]
fenjalien: `apply-shape-transform` would work but I prefer just `transform-shape`

src/draw/shapes.typ
fenjalien: I disagree that the transformation would be too "hidden", the `transform` argument would be explicitly passed and would always happen. I think having and repeating some code that affects marks outside of a mark function is not good.

src/styles.typ
fenjalien: Yes thats true

manual.typ
```diff
@@ -278,6 +278,7 @@ line(..c)
 
 #doc-style.show-parameter-block("shorten-to", ("integer", "auto", "none"), [Which mark to shorten the path to when multiple marks are given. `auto` will shorten to the last mark, `none` will shorten to the first mark (effectively disabling path shortening). An integer can be given to select the mark's index.])
 
+#doc-style.show-parameter-block("shape-transform", "bool", [If false, place marks post path transformation, which results in marks not affected by the current transformation.], default: true)
```
fenjalien, line 281: I think this description should have the effect first then why: "When false marks will not be streched/affected by the current transformation, marks will be placed after the path is transformed."

I do also think "transform-shape" would be a better name...

src/draw/shapes.typ
```diff
@@ -313,19 +313,18 @@
     )
 
     if mark_.check-mark(style.mark) {
-      let (marks, segments) = mark_.place-marks-along-path(ctx, style.mark, drawables.segments)
+      let (marks, segments) = mark_.place-marks-along-path(ctx, style.mark, transform, drawables)
       drawables.segments = segments
       drawables = (drawables,) +  marks
```
fenjalien, line 318: If drawables are being passed completely, why not add the marks to the drawables and return only the drawables?
```
drawables = mark_.place-marks-along-path(ctx, style.mark, transform, drawables)
```

src/drawable.typ
```diff
@@ -3,6 +3,19 @@
...
+    return segments
+  }
+
+  assert(type(segments) == array,
+    message: "Expected segment array, got " + repr(segments))
+  return segments.map(segment => {
+    let (kind, ..pts) = segment
+    return (kind,) + util.apply-transform(transform, ..pts)
+  })
+}
```
fenjalien, line 17: Would this be better in path-utils?

Because this function is used in a loop in `apply-transform`, I would move lines 7 to 12 to another function which then calls this one. So the transform if statement isn't done and segments assert is not done every loop.
```
let apply-transform-segments-unchecked(transform, segments) = {...}
let apply-transform-segments-checked(transform, segments) = {
  if transform ...
  assert(...)
  apply-transform-unchecked(...)
}

let apply-transform(...) = {
  ...
  if transform ...
  for drawable in drawables {
    ...
    apply-transform-segments-unchecked(transform, drawables.segments)
  }
```
Although at this point it starts to become a mess...

src/mark.typ
```diff
@@ -289,5 +296,12 @@
       snap-to: snap-to)
   }
 
+  // If not transformed pre mark placement,
+  // transform everything after mark placement.
+  if not is-transformed {
+    segments = drawable.apply-transform-segments(transform, segments)
+    drawables = drawable.apply-transform(transform, drawables)
+  }
```
fenjalien, line 304: As said in `draw/shapes.typ` if this pushes the mark onto the path drawables, you would then only need a single `apply-transform` call and no `apply-transform-segments` call.
You would also be able to just use `apply-transform` on line 264.

manual.typ
johannes-wolf: Changed.

src/drawable.typ
johannes-wolf: Removed.

fenjalien (approved):

PR #494, opened 2024-02-11
plugin: A plug-in API

johannes-wolf:
### Plug-Ins
This is a PoC of a plug-in API for cetz.

Other packages can register themselves as a plugin by providing a plug-in dictionary, which can contain callbacks to initialize a canvas.

Currently, a single callback is supported:
- `init: () => element`: Can return canvas elements that get prefixed to each canvas body

### Example
```
#let my-plugin = (
  init: () => {
    draw.set-style(stroke: blue)
    draw.content((0, 1), [Demo])
  }
)
#plugin.register(my-plugin)

#canvas({
  import draw: *
  // Renders the plug-ins content and sets the stroke to blue
  line((0,0), (1,1))
})

#canvas({
  // This just renders the plug-ins content
})
```

Because of how Typst `state` works, care has to be taken when relaying on custom context data of a plug-in.

### Alternative
We could also not rely on `state` and just accept a `plug-ins: (...)` list in `canvas`. This would have the benefit of being sure plug-ins are ready when the canvas gets drawn. This should be more performant, as with the state based solution, each canvas (could) get(s) drawn multiple times with different state values. Using `state.final(loc)` did not help.

### DIFF
--- a/manual.typ
+++ b/manual.typ
@@ -904,6 +904,33 @@ All path decoration functions support the following style keys:
 
 #doc-style.parse-show-module("/src/lib/decorations/path.typ")
 
+=== Plug-Ins
+CeTZ supports plug-ins that get automatically called for each canvas.
+A plug-in can modify the initial state of a canvas by providing any of the
+callbacks `init` or `draw`.
+
+#doc-style.parse-show-module("/src/plugin.typ")
+
+==== Example
+```typsc
+plugin.register((
+  init: ctx => {
+    ctx.my-plugin-data = [Demo]
+    return ctx
+  },
+  draw: ctx => {
+    cetz.draw.set-style(stroke: blue)
+    cetz.draw.content((0,0), ctx.my-plugin-data)
+  }
+))
+...
+canvas({
+  // All plug-in's draw functions get automatically injected here
+  import draw: *
+  line((0,0), (1,1))
+})
+```
+
 ==== Styling
 
 ===== Default `brace` Style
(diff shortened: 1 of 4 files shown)


fenjalien:
I need to get on with work so this is going to be a "quick" ramble that might not make sense:

I agree with wrapping the canvas with `location` to give the user as much context for the canvas. However, what benefit does using state, or even passing it to the canvas, give over recommending that the integrations have a `init-` type function that the user should call at the start of each canvas?
If the user does want to have the plugin for every canvas, and has many, they could write a wrapper function for the canvas. The integrations could evern provide one, although this may prevent multiple integrations from being used together.

Using state means it effects all canvases after it, so there should be a way to remove plugins from the state. 

The idea of passing a dict of functions with keys so they get called at particular points seems very pre-0.2, could you give an example of additional keys that could be added?

johannes-wolf:
> I need to get on with work so this is going to be a "quick" ramble that might not make sense:
> 
> I agree with wrapping the canvas with `location` to give the user as much context for the canvas. However, what benefit does using state, or even passing it to the canvas, give over recommending that the integrations have a `init-` type function that the user should call at the start of each canvas? If the user does want to have the plugin for every canvas, and has many, they could write a wrapper function for the canvas. The integrations could evern provide one, although this may prevent multiple integrations from being used together.
> 
> Using state means it effects all canvases after it, so there should be a way to remove plugins from the state.
> 
> The idea of passing a dict of functions with keys so they get called at particular points seems very pre-0.2, could you give an example of additional keys that could be added?

Yes, the state mechanism is not a good idea, because of the multiple canvas invcations. The only nice thing is that you do not need any wrapper functions etc. I will close this PR. I guess having no special API for plug-ins but doing it like in the Polylux binding PoC PR is the way to go.

PR #502, opened 2024-02-13
canvas: Change to rh coordinate system

johannes-wolf:
This is a breaking change for code that uses 3D coordinates!

Right hand sided coordinate systems are more common. Cetz currently uses lhs because TikZ does so by default. But TikZ also makes it more easy to change the system by having x, y and z arguments to the tikzpicture/canvas.

Question: Should we add `x`, `y`, and `z` arguments to `canvas` to make setting up a custom
coordinate system simpler?

An alternative would be providing helper functions to set up axes:
```typst
set-axes(x, y, z) // Signature
set-axes(1, -1, -.5) // Set length relative values (default xyz system)
set-axes(2cm, .5cm)  // Set absolute lengths (default xyz system)
set-axes((0,1), (1,0), (.5,.5)) // Set custom axis vectors
```

Such function would override the current transformation matrix.
The proposed x, y and z arguments would be able to do the same (accept number or vector).

---

The proposed arguments or functions are not part of this PR but a proposal for a follow-up PR.


fenjalien:
I don't have any particular opinion about this as I haven't and don't have the need to use 3D. However, I would like to get feedback from users about this before merging as I assume it could be a large breaking change (although easily fixed just annoying). A thread on the discord server would work :)

KronosTheLate:
Before, z was sheared by 0.5. Now, it is instead scaled by 0.5. I am not sure what the effect of that difference is. Other than that, the sign is swapped, and this looks good to me!

johannes-wolf:
3D users I know of that already use a rhs system by manually transforming coordinates:
- https://typst.app/project/pJEyBsAL5xquEC6fqfyvZ7 (by loewe1000)

johannes-wolf:
Ping.

knuesel:
For what it's worth I did a lot of 3D plots in TikZ and first thing to do is always using the keys `x`, `y`, `z` to define a right-handed coordinate system with z looking up. (I'm still trying to figure out how to do that in CeTZ :-) ). Having right-handed by default would make a lot of sense for me.

By the way is the default really left-handed in TikZ?  I have no memory of that and trying this:

```
\documentclass{article}
\usepackage{tikz}
\begin{document}
\begin{tikzpicture}
\draw (0,0,0) -- (1,0,0) node {x};
\draw (0,0,0) -- (0,1,0) node {y};
\draw (0,0,0) -- (0,0,1) node {z};
\end{tikzpicture}
\end{document}
```

gives the following which looks right-handed:

image

johannes-wolf:
You would do that using `scale(x: , y:, z: )`

KronosTheLate:
I was honestly very surprised to see the default being a left handed coordinate system. Throughout my 5 years engineering education I have not once used a left handed one, nor seen anyone else do it. 

Are there any fields in which it is not just implied that coordinate systems are right handed unless otherwise specified?

fenjalien:
Okay I've seen a few comments on discord as well, I'm happy for it to be merged once its been updated

tests/projection/default/test.typ
fenjalien, line 1: Add page auto size

johannes-wolf: Done.

fenjalien (approved):

PR #503, opened 2024-02-13
Don't add center or default anchors to groups with no bounds

fenjalien:


johannes-wolf (approved):

johannes-wolf:
Please add a test-case for this issue.

PR #504, opened 2024-02-13
projection: Add orthographic projection utilities

johannes-wolf:
This PR adds some helper functions for setting up orthographic projection.

Three functions for drawing on different planes got added: `on-xy`, `on-xz` and `on-yz`.
These allow elements which do not allow to be used with different z-coordinates to be drawn in 3D space without having to use rotations etc.

The new `ortho` function sets-up an isometric projection by default:
grafik

There is no z-sorting (yet?).

<details>
  <summary>Code</summary>

  ```typst
#import "src/lib.typ" as cetz

#cetz.canvas({
  import cetz.draw: *

  let cube() = {
    set-style(
      line: (mark: (end: ">", scale: .3)))

    scale(3)
    line((-1,0,0), (1,0,0))
    line((0,-1,0), (0,1,0))
    line((0,0,-1), (0,0,1))

    set-style(
      stroke: (paint: black, thickness: 1pt, join: "round"),
      fill: rgb(0,0,255,50))

    scale(.35)
    on-xy(z: -1, rect((-1,-1), (1,1)))
    on-xy(z: +1, rect((-1,-1), (1,1)))
    on-xz(y: -1, rect((-1,-1), (1,1)))
    on-xz(y: +1, rect((-1,-1), (1,1)))
    on-yz(x: -1, rect((-1,-1), (1,1)))
    on-yz(x: +1, rect((-1,-1), (1,1)))
  }

  ortho(name: "ortho", {
    cube()
  })

  set-origin("ortho.east")
  group(name: "ortho", anchor: "west", ortho(x: 20deg, {
    cube()
  }))

  set-origin("ortho.east")
  group(name: "ortho", anchor: "west", ortho(y: 20deg, {
    cube()
  }))
})

#cetz.canvas({
  import cetz.draw: *

  ortho({
    scale(2)
    on-xy(z: -1.1, {
      rect((-1,-1), (1,1))
      translate((-.9, -.9))
      cetz.plot.plot(size: (1.8, 1.8), x-tick-step: none, y-tick-step: none, x-label: none, y-label: none, {
        cetz.plot.add(domain: (-2, 2), x => x * x * x)
        cetz.plot.add-contour(
          (x, y) => 2 - (x - 1) * (y - 1),
          fill: true,
          x-domain: (-2, 2),
          y-domain: (-10, 11))
      })
    })
    on-xz(y: -1.1, {
      rect((-1,-1), (1,1))
      grid((-.9, -.9), (.9, .9), step: .2)
    })
    group(name: "g3", {
      on-yz(x: -1.1, {
        rect((-1,-1), (1,1))
        for r in range(1, 10) {
          circle((0,0), radius: .1 * r)
        }
      })
    })
  }, name: "gO")
})

  ```

</details>


KronosTheLate:
That looks neat! Would it be possible to allow helper-functions/krwargs to make ticks and axis-labels? E.g. `ortho(content, axis-labels=true)` for default label style and position, and `ortho(content, axis-labels=(x, y, z))`, with x y z being either strongs or content. And similarly a kwarg `tick-spacing` which is a number applied to all axis, or a vector of number for finer control.

This looks nice though. Easy to use and rotate round any axis. However, would this be much different from defining your object, and rhen rotating the entire thing? I am not sure what `ortho` adds that is not east enough already.

johannes-wolf:
> That looks neat! Would it be possible to allow helper-functions/krwargs to make ticks and axis-labels? E.g. `ortho(content, axis-labels=true)` for default label style and position, and `ortho(content, axis-labels=(x, y, z))`, with x y z being either strongs or content. And similarly a kwarg `tick-spacing` which is a number applied to all axis, or a vector of number for finer control.

Adding axes with ticks should be a separate Element.

> This looks nice though. Easy to use and rotate round any axis. However, would this be much different from defining your object, and rhen rotating the entire thing? I am not sure what `ortho` adds that is not east enough already.

- It gives an environment for 3D, we could even sort by camera z-order insides `ortho`.
- I do not really like the default z-shearing transformation, we could remove it
- Otho outputs "flat" coordinates with no z-value (outsides the environment), which is useful with anchors and other elements.

KronosTheLate:
Alright, I can see how `on-xy` and friends, plus `ortho` would be useful. I am unable to really check the pros and cons of this implementation in my mind, but like it, and do not immediately have any reservations about it ^_^ I would also say that tagging breaking releases in quick succession due to trying things out if not a big problem as I see it, given the maturity of both Typst and all its packages. So I feel like it makes sense to give this a go, and to not be afraid to change it if usage reveals issues.

src/draw/projection.typ
```diff
@@ -0,0 +1,115 @@
...
+#let ortho-matrix(x, y, z) = matrix.mul-mat(
+  matrix.mul-mat(
+    matrix.mul-mat(
+      matrix.ident(),
+      matrix.transform-rotate-x(x)
+    ),
+    matrix.transform-rotate-y(y)
+  ),
+  matrix.transform-rotate-z(z)
+)
```
fenjalien, line 22: `matrix.mul-mat` can take separate matrices as positional arguments and mulitply them together like this.

src/draw/projection.typ
fenjalien, line 1: Would it make more sense to have this under a lib instead of draw?

Could you also add comments to describe what is happening in the code and for each function.

src/draw/projection.typ
```diff
@@ -0,0 +1,115 @@
...
+
+
+#let _projection(body, view-matrix, projection-matrix, reset-transform: false) = {
+  (ctx => {
+    let transform = ctx.transform
+    ctx.transform = matrix.mul-mat(projection-matrix, view-matrix)
+    if not reset-transform {
+      ctx.transform = matrix.mul-mat(transform, ctx.transform)
+    }
+    let (ctx, drawables, bounds) = process.many(ctx, body)
```
fenjalien, line 32: Use the new `util.resolve-body` function

src/draw/projection.typ
```diff
@@ -0,0 +1,115 @@
...
+        let (kind, ..pts) = s
+        return (kind, ..pts.map(pt => {
+          fn(pt)
+        }))
+      })
+      if d.type == "content" {
+        d.pos = fn(d.pos)
+      }
+      return d
+    })
```
fenjalien, line 61: Is this really the best way to do this? Is there no transform that could be used instead?

Could you clear up the segments map as well maybe:
```
(s.slice(0, 1), ..s.slice(1).map(...))
```

fenjalien (commented):
This is really cool.

Standard extra stuff, manual, changelog, tests are still needed

src/draw/projection.typ
johannes-wolf: What do you mean by "clear up"?
You should be able to get the same result using rotation + scaling + translation.

src/draw/projection.typ
johannes-wolf: Hmm, I think this is very basic functionality. Since we support 3D coordinates, we should also have some utility functions to help to use them.

src/draw/projection.typ
fenjalien: It looks difficult to read, so tidy it up so it reads better:
```
d.segments.map(s => { (s.slice(0, 1), ..s.slice(1).map(fn)) })
```

src/draw/projection.typ
fenjalien: Okay I guess that makes sense.

Can you add some more comments in the `_projection`, `apply-transform-function`, `ortho`, and one of the `on-*` functions to describe whats happening with the transformation matrices

johannes-wolf: I've added comments explanining how the vertices are transformed and added exapmles for each function.

fenjalien (approved):
Update changelog first

johannes-wolf:
@fenjalien Do we want to add this to 0.2.2? This is a non-breaking change.

fenjalien:
I don't see why not :)

PR #505, opened 2024-02-14
canvas: Change default transform to remove z

johannes-wolf:
Change the default transformation matrix to have z coordinates of zero after transformation.
This is actually a fix, as the final 2D projection should not have a z-value other than zero.

With this change, anchors in 3D space become 2D and thus elements like rect, Bézier or Catmull, … will correctly work when using them with anchors to 3D elements. Intersection between 3D elements will also work (in 2D space).

See: #502 and #493.


src/canvas.typ
```diff
@@ -42,10 +42,11 @@
     em-size: measure(box(width: 1em, height: 1em), st),
     style: styles.default,
     // Current transform
-    transform: matrix.mul-mat(
-      matrix.transform-shear-z(.5),
-      matrix.transform-scale((x: 1, y: -1, z: 1)),
-    ),
+    transform:
+      ((1, 0,+.5, 0),
```
johannes-wolf, line 46: I guess hand writing the matrix, instead of using 2 multiplications, is more clear.

johannes-wolf:
With this PR you can get TikZ like marks on 3D lines:
```typst
  set-style(mark: (end: ">", transform-shape: false))
  line((0,0), (1,0))
  line((0,0), (0,1))
  line((0,0), (0,0,1))
```
grafik

fenjalien (commented):
- [ ] Add a comment to the default transform matrix to say what transforms it has e.g. shear then scale... xyz
- [ ] Check if the manual says anything about the default transform, update it if so
- [x] Update changelog

johannes-wolf:
> * [x]  Add a comment to the default transform matrix to say what transforms it has e.g. shear then scale... xyz
> 
>     * [x]  Check if the manual says anything about the default transform, update it if so
> 
>     * [x]  Update changelog

Added a line to the changelog and documentation about the default matrix.

fenjalien (approved):

PR #513, opened 2024-02-23
chart: Fix piechart clockwise style

johannes-wolf:
For piecharts with `clockwise = true`, the slice-styles are in reverse order.
This fixes the order of the styles, which were still ccw.


fenjalien (approved):
Update the changelog first :)

PR #516, opened 2024-02-27
Data labels

rousbound:
Implementing 'data-labels'. 

That is, the y values floating next/above each bar.

graf1
graf2

### DIFF
--- a/src/lib/chart/columnchart.typ
+++ b/src/lib/chart/columnchart.typ
@@ -5,8 +5,10 @@
 
 #let columnchart-default-style = (
   axes: (tick: (length: 0), grid: (stroke: (dash: "dotted"))),
-  bar-width: .8,
-  x-inset: 1,
+  bar-width: .9,
+  x-inset: 0.6,
+  data-label: (offset: 0.20, text-size: 8pt),
+  bars-padding: 0.02
 )
 
 /// Draw a column chart. A column chart is a chart that represents data with
--- a/src/lib/plot/bar.typ
+++ b/src/lib/plot/bar.typ
@@ -66,9 +66,7 @@
 
 
       draw.group(ctx => {
-        let right-padding = right - ctx.style.bars-padding
-        let left-padding = left + ctx.style.bars-padding
-        draw.rect((left-padding, y-min), (right-padding, y-max))
+        draw.rect((left, y-min), (right, y-max))
         if ctx.style.data-label != none {
           let offset = ctx.style.data-label.at("offset")
           let size = ctx.style.data-label.at("text-size")
--- a/src/lib/plot/bar.typ
+++ b/src/lib/plot/bar.typ
@@ -187,6 +187,5 @@
     plot-legend-preview: self => {
       draw.rect((0,0), (1,1), ..self.style)
     }
-    ),
-  )
+  ))
 }
--- a/src/lib/plot/bar.typ
+++ b/src/lib/plot/bar.typ
@@ -52,11 +52,11 @@
     let right = left + w
     let width = (right - left) / len
 
+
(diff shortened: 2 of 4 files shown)


johannes-wolf:
I will take a look at this at the end of next week.

src/lib/plot/bar.typ
```diff
@@ -52,19 +52,30 @@
...
       y-max = calc.min(y-max, y-axis.max)
 
-      draw.rect((left, y-min), (right, y-max))
+
+      draw.group(ctx => {
+        draw.rect((left, y-min), (right, y-max))
+        if ctx.style.data-label != none {
+          let offset = ctx.style.data-label.at("offset")
+          let size = ctx.style.data-label.at("text-size")
+          let data_label = text(size:size)[#y-max]
```
johannes-wolf, line 73: The label itself should come from a callback, so the user can format it etc.
To do this, add a `data-label` or `bar-label` (?) argument to `add-bar`.
We could allow some special values like `"max"` to auto use a callback that shows the maximum value.

Such callback should receive the "current" bar as argument.

I do not like having a font-size as style attribute. I prefer the user setting the size via the label callback.

src/lib/plot/bar.typ
```diff
@@ -52,19 +52,30 @@
...
       right = calc.min(right, x-axis.max)
       y-min = calc.max(y-min, y-axis.min)
       y-max = calc.min(y-max, y-axis.max)
 
-      draw.rect((left, y-min), (right, y-max))
+
+      draw.group(ctx => {
+        draw.rect((left, y-min), (right, y-max))
+        if ctx.style.data-label != none {
+          let offset = ctx.style.data-label.at("offset")
```
johannes-wolf, line 71: Style data must come from `self.style`. This allows setting per-bar/plot style data.

src/lib/plot/bar.typ
```diff
@@ -52,19 +52,30 @@
...
+          let size = ctx.style.data-label.at("text-size")
+          let data_label = text(size:size)[#y-max]
+          let anchor = if y-axis.horizontal {"west"} else {"south"}
+          draw.content((rel: (0, offset), to: ((left + right) / 2, y-max)), anchor:"south", data_label)
```
johannes-wolf, line 75: Prefer giving a name to the bar rect and using the `"north"/"south"` or `"east"/"west"` anchors, depending on the orientation _and_ value.

Offset must work for horizontal and vertical bars!

src/styles.typ
```diff
@@ -15,6 +15,7 @@
   //   - Array: (y, x), (top, y, bottom), (top, right, bottom, left)
   //   - Dictionary: (top:, right:, bottom:, left:)
   padding: none,
+  data-label: none,
```
johannes-wolf, line 18: Remove this line.

src/lib/chart/columnchart.typ
```diff
@@ -5,8 +5,10 @@
 
 #let columnchart-default-style = (
   axes: (tick: (length: 0), grid: (stroke: (dash: "dotted"))),
-  bar-width: .8,
-  x-inset: 1,
+  bar-width: .9,
```
johannes-wolf, line 8: Shouldn't bars and columns have the same height/width?

johannes-wolf (changes requested):
Thank you for your PR!

There is still one big problem with adding `content` inside plots, which is yet to be solved (via a separate PR): Calculating scaling for scaled and fixed size elements inside the scaled plot environment; so this PR might result in overlapping data labels (which is ok for now).

johannes-wolf:
Closing because `cetz-plot` will be/got moved out of `cetz`.
You can re-open this PR in the new repo.

PR #526, opened 2024-03-15
Apply Typst 11 Changes

johannes-wolf:
This PR replaces `style` with the new `context` keyword.


src/canvas.typ
```diff
@@ -34,18 +34,17 @@
...
     message: "Canvas length must be != 0!")
 
   let ctx = (
-    typst-style: st,
     length: length,
     debug: debug,
     // Previous element position & bbox
     prev: (pt: (0, 0, 0)),
-    em-size: measure(box(width: 1em, height: 1em), st),
+    em-size: 1em.to-absolute(),
```
fenjalien, line 47: Do we need to have `em-size` in ctx anymore?

src/util.typ
fenjalien, line 1: Can `reslove-number` be improved as well? We currently `repr` the number to check if its an em unit which we shouldn't have to do anymore

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - uses: yusancky/setup-typst@v2
         id: setup-typst
         with:
-          version: 'v0.10.0'
+          version: 'v0.11.0'
```
fenjalien, line 39: Theres a community action (https://github.com/typst-community/setup-typst) that we should use instead

fenjalien (approved):

PR #531, opened 2024-03-16
matrix: Enable rounding for matrix multiplication

johannes-wolf:
Round multiplied cells in `mul-mat`.


fenjalien (commented):
Looks great but its "precision" with one "s"

PR #535, opened 2024-03-20
clipping: Add path clipping support

johannes-wolf:
This PR implements boolean path operations. The implementation is handled by a C++ library: contourklip.


fenjalien (commented):
Okay I've come around to the idea of keeping plugins in this repository. This is mostly due to the fact that I can't think of any alternatives that would cause more hassle and problems than they're worth.

I would like to propose a slight restructuring though. Put the plugin in a dedicated folder `plugins/clipping/` and call it `clipping.wasm` instead of `cetz.wasm`. 

The draw and lib folders probably need a restructure as well but that can be left to another PR.

johannes-wolf:
We can not use contourklip because of it being GPL.

Issue #465, opened 2024-01-17
Allow decoration to be set with style keys.


fenjalien:
You currently have to wrap the path that you want to decorate with another decoration function which effectively modifies the output drawables of the path. Instead (or alongside) I would prefer the option to set a path's decoration through a style key:
```
line((0,0), (1,1), decoration: "zigzag")
```
Styling options for the decoration could be given further as a dictionary similar to how mark styling works.

johannes-wolf:
Idea we've discussed:
Allow to register path modifiers at the context object. So libraries can add their custom path modifiers, like with user definable (`define-mark`) marks.

johannes-wolf:
*API idea*
Elements have a `modifier:` style key. A modifier can be the name of a modifier function or a dictionary containing the name + style arguments. A user can pass a list of modifiers to an element. All modifiers share some common keys like `begin` and `end` to denote the sup section of a path to apply the modifier to.

The modifier API replaces the current path decoration lib.

There will be a separate `apply-modifier(.., body)` function to apply a modifier to a body.

fenjalien:
Considering what I've detailed in #467 I don't agree. Restricting modifiers to be like this feels like a waste of potential. Decorations should really be a subset of modifiers.

Issue #467, opened 2024-01-17
proposal: Add drawable modifiers


fenjalien:
As a way to generalise such things as marks and decorations we should have element modifiers. It would come in the form of a single function that takes an element's drawables, ctx and the element's styling (and possibly the name and type of the element). Then return the modified drawables which can then be transformed and anchors applied to.

The function would then call modifier functions in ctx, depending on the state of the style dict (and other options like the name or type), to operate on the drawables.

This would:
  - Replace how marks are done, and may change the outcome of #465.
  - Allow new and custom user features that modify an element's path without having to wrap the draw function.

PR #541, opened 2024-03-30
shapes: Use decoration style

johannes-wolf:
Implements #465.

This PR is a WIP implementation on how to implement decorations as style option for all (but grid and mark) elements.

The idea:
- The context has a `decorations` dictionary that can be modified to add custom decorations.
  This allows for a function like `register-decoration(...)` for plug-ins (same plan as with custom marks; not added yet).
- The user can either use a decoration name (dict) or pass a function to `decoration: ..` for elements. Element paths get decorated _before_ adding marks, which allows for styled paths to have marks.
- You can still use decoration-functions via composition.
- You can apply multiple decorations in sequence
- Decorations can be applied on a subsection of a path (start, end)
- This API will replace the current decoration lib
- It works without destroying marks

Demo of the `wave` and `linearize` modifiers:
grafik

### DIFF
--- a/src/draw/shapes.typ
+++ b/src/draw/shapes.typ
@@ -20,6 +20,45 @@
 #import "styling.typ": *
 #import "grouping.typ": *
 
+
+// Apply path decoration function fn to drawables
+// and return a single drawable.
+#let _apply-decoration-fn(ctx, style, drawables, fn) = {
+  if fn != none {
+    ctx.transform = matrix.ident()
+    let (drawables: drawables, ..) = process.many(ctx, (fn)((ctx => {
+      return (ctx: ctx, drawables: if type(drawables) != array { (drawables,) } else { drawables })
+    },), ..style, decoration: none))
+
+    assert.eq(drawables.len(), 1,
+      message: "Path decoration function must return a single element")
+    return drawables.first()
+  }
+  return drawables
+}
+
+// Apply path decoration to drawables
+#let _apply-decoration(ctx, style, drawables) = {
+  let deco = style.at("decoration", default: none)
+
+  if deco != none {
+    import "/src/lib/decorations.typ": wave, zigzag, coil
+    assert(type(deco) in (str, function),
+      message: "Decoration must be of type string or function")
+
+    let fn = if type(deco) == str {
+      ctx.decorations.at(deco)
+    } else if type(deco) == function {
+      deco
+    }
+
+    assert(fn != none,
+      message: "Decoration function is none")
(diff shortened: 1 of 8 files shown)


src/draw/shapes.typ
```diff
@@ -20,6 +20,45 @@
...
+    } else if type(deco) == function {
+      deco
+    }
+
+    assert(fn != none,
+      message: "Decoration function is none")
+    return _apply-decoration-fn(ctx, style, drawables, fn)
+  }
+  return drawables
+}
```
fenjalien, line 60: Move these functions to `lib/decorations.typ` even if they should only be used with shape elements they can still be used to apply decorations to other elements.

fenjalien (commented):
We'll need to write a specification of what a decoration function should expect to receive and what it is expected to return.

What do you think of blocking this until we work on modifiers?

src/draw/shapes.typ
johannes-wolf: Yes. Do we want to have some kind of function to apply modifiers to a path post construction? Some generic function that accepts a modifier + path. The existing deco functions would then also use that. We should move the implementation out of lib, as functions from lib should not be used by core cetz.

A modifier function should be specified as a function taking a context, styles and a single drawable and outputs a single drawable.

fenjalien: We should discuss this further in #467

johannes-wolf:
@fenjalien I've updated the API. This new modifier implementation is meant to completely replace the `decortaions.path` library and move path modifiers to cetz core.
I think `decoration:` should be named `modifier:`.

The PR is still very WIP but should give a first idea. It is a better implementation than what `decorations.path` currently provides.

johannes-wolf:
Rebased.
How do we want to proceed with this?
I am open to change the name from `modifier` to something else, although I prefer `modifier` or `decoration` over `show-rule`, as it is more explanatory.

If we merge this, we can implement marks on top of this. I prefer _not_ touching the mark implementation before we agree on a modifier/decoration implementation.

If we want to support outputting extra geometry from modifiers, we should discuss how to do that. The current interface does not allow this. To add support, the `apply-*` function must know if the returned drawable is part of a continuous path or if it is some “standalone” shape.

fenjalien:
I'm happy to continue using `modifier`, but I think using `show-*` would help to draw a parallel to Typst's `show` rules. I don't like `decoration`.

I would really like to get on with this but I want to get the docs website up properly and I'm struggling to find time for it.

PR #542, opened 2024-03-30
Change how `copy-anchors` works

### LINKED ISSUE #538: Panics when copying anchors from a group with a named element with no default anchor
CeTZ 0.2.1 (Typst 0.11.0) panics with the following code
```
group(name: "a", {
  group(name: "b", {
    line((), (1,1), name: "l")
  })
  copy-anchors("b")
})
```
`"Anchor 'default' not in anchors () for element 'l'"`

The behaviour how `copy-anchors` acutally copies the anchors should change, I'm not sure how right now though.

fenjalien:
Fixes #538.

It behaves functionally the same as before but instead of calculating all of the anchors to be copied within the function, it wraps each anchor to be copied in another function to act as a proxy. This allows nested anchors/access to copied elements.

The only downside I see is another function on the call stack...


johannes-wolf (approved):
LGTM.

Issue #472, opened 2024-01-18
Documentation Website


fenjalien:
Instead of (or as well as) writing a pdf manual we should have documentation on a website. Ideally this would be hosted through github pages and be generated from the current docstrings we have.

Some options to use:
- [typst-book](https://github.com/Myriad-Dreamin/typst-book)
  - @johannes-wolf has tested it but it does not seem stable enough
- [mdBook](https://rust-lang.github.io/mdBook/)
  - From a quick glance this might be the best option if we mix it with pandoc. It allows the running of preprocessors so examples could be generated automatically.
- [readthedocs](https://docs.readthedocs.io/en/stable/)

johannes-wolf:
I guess we should give typst-book another try, as the last test was some time ago and as far as I know the performance has improved.

Enivex:
Maybe @Myriad-Dreamin can give some input

fenjalien:
Probably should have updated this sooner but a documentation site is setup! Well at least the infrastructure/site is, the actual content and stuff still need to be moved over to it.

It uses mdbook with a custom preprocessor that adds rendered examples, type pills like the Typst docs, and parameter defintions like we use in the manual.

You can find it here: https://cetz-package.github.io/docs/

PR #548, opened 2024-04-05
doc: Add default value of `shorten-to` to the docs

### LINKED ISSUE #543: Default value of `shorten-to`
Hi! Thanks for the amazing package!

Just a small bug/discrepancy I noticed between docs and implementation of marks. In the docs, it's stated that the default value for `shorten-to` parameter is `none` (which I believe is the intended implementation), however in [here](https://github.com/cetz-package/cetz/blob/3562d65ee239342507039c2dbdd599231428cbf8/src/mark.typ#L32) the default value is set to `auto`.

MWE:

```
#cetz.canvas({
  import cetz.draw: *
  line((0, 1), (5, 1))
  line((0, 0), (5, 0), mark: (symbol: ">", pos: 50%))
})
```

image

johannes-wolf:
Fixes #543


fenjalien:
I don't think this PR would be necessary with #472

johannes-wolf:
> I don't think this PR would be necessary with #472

Let's merge it anyways.

PR #551, opened 2024-04-09
Fix a few small punctuation errors in the manual

kdkasad:
As the title says, this fixes a few small punctuation errors in the manual.
None of these affect the readability of the document; they're all for cosmetic purposes.

Changes:
 - Change "Typst's function's" to "Typst's functions"
 - Remove double-punctuation at the end of one sentence ("!." to "!")
 - Move commas inside quotation marks (see line 108).  
   This one is debatable, as the quotation marks in this case are being used to refer to exact values, so I can understand why one might want to keep the commas outside the marks.


johannes-wolf (approved):

johannes-wolf:
Thank you! We are now also working on a web documentation, see: https://github.com/cetz-package/docs

kdkasad:
> Thank you! We are now also working on a web documentation, see: https://github.com/cetz-package/docs

Good to know! I've never actually used CeTZ, so I can't help much with the docs. I just noticed the typos in this PR when I was reading the manual to learn CeTZ.

PR #559, opened 2024-04-14
mark: Fix mark position + allow angle as coordinate

### LINKED ISSUE #557: Bug in Mark positioning
According to documentation `Mark` have 2 coordinates `from` and `to` and it should be placed at `from` and point in direction of `to` however, right now it is placed at `to` instead of `from`. Looks like virtual path is created between `from` and `to` and mark is placed there but instead at start it is placed at end of that path. Current code looks like this:

code from function: `#let mark(from, to, ..style) = {`

```
if style.end == none {
      style.end = style.symbol
}
style.start = none
style.symbol = none
```

This looks like bug, `style.start` should be used not `style.end`

johannes-wolf:
Fixes #557.

This PR fixes the position bug and adds the option to pass an angle as second coordinate to `mark`.


src/draw/shapes.typ
```diff
@@ -442,6 +442,13 @@
   )
   
   let style = style.named()
+
+  if type(to) == angle {
+    // Construct a coordinate pointing (+1, 0) away from
+    // `from`, rotated by the angle given.
+    to = (from, 1, to, (rel: (1,0), to: from))
```
fenjalien, line 449: would it be better to use polar and relative? `(rel: (1, to), to: from))`

src/draw/shapes.typ
```diff
@@ -454,7 +461,10 @@
     style.start = none
     style.symbol = none
 
-    let drawables = drawable.path((path-util.line-segment(pts),))
+    let (from, to) = (..pts)
+    let dir = vector.sub(to, from)
+
+    let drawables = drawable.path((path-util.line-segment((vector.sub(from, dir), from)),))
```
fenjalien, line 467: This looks very confusing, whats wrong with `line-segment(from, vector.neg(to))`?

johannes-wolf: How should that work?
The last argument has to be `from`, because that is where the end mark should be drawn.

I've reworked it, but we have to mirror `to` at `from` to make it work correctly.

fenjalien (approved):

PR #562, opened 2024-04-21
ci: Remove typst-test tag

johannes-wolf:


fenjalien:
shouldn't we match the new branch now as well?

johannes-wolf:
> shouldn't we match the new branch now as well?

@tingerrr: What should be used: master or the semi-stable branch?

tingerrr:
Yes indeed, I'm not sure when or why I decided to make it a tag. I think, when I tried it out, it caused the CI scripts to download redownload typst-test even when the branch was not moved. But the tag approach caused more churn during development because of how git fetches them, so I wanted to move away from this.

PR #583, opened 2024-05-10
plot: Add some type checking assertions

johannes-wolf:
Add some type checking assertions to `plot.plot`.


fenjalien (approved):
I suppose these checks are fair but could they be mitigated by improving the documentation?

johannes-wolf:
> I suppose these checks are fair but could they be mitigated by improving the documentation?

I think it is quite clear that those values should be dictionaries (or functions), as they are documented as such. These checks are because someone passed other values to those fields and the resulting error message was not that helpful.

PR #584, opened 2024-05-11
mark: Add anchor style to shift marks

### LINKED ISSUE #578: Mark positioning
In currect CeTZ implementation when we place mark at point `(x,y)` it is tip of mark that will occupy that position. In case on marks that are large or long it creates some issues to center in properly in the middle of line, because in most cases we would like to position mark in such way that graphical center of mark will occupy `(x,y)`. Image below. Top part represent current mark positioning behaviour, bottom part my tweaks to place mark relative to is graphical center. Second one (mark shifted by `mark_width/2`) seems more intuitive I think.

Screenshot 2024-05-02 095941

johannes-wolf:
@fenjalien Not sure about the name `origin`. The idea is to have a style key that shifts the
mark origin along its (local) x-axis: 0% tip, 100% base.

- [x] Add documentation
- [x] Add changelog

Fixes #578.


fenjalien:
Would it be easier to add anchors to the mark function? "tip", "center" and "inset" (or "base") could work

johannes-wolf:
> Would it be easier to add anchors to the mark function? "tip", "center" and "inset" (or "base") could work

As an addition? Because this feature is also useful when placing a mark on a path, e.g. in the center.

fenjalien:
No, instead of using the `origin` styling parameter. Anchors would functionally do the same thing but with predefined positions. Given the normal size of marks I don't think you would want such fine control it.

arkadiuszwojcik:
@fenjalien in such case how it would look like if I would like to use mark center as placement origin?

johannes-wolf:
`mark(..., anchor: "center")`, I guess.

johannes-wolf:
Changed. You can now specify one of the anchors `tip`, `center` and `base` to place the mark at. This also works with marks on paths.

src/styles.typ
```diff
@@ -49,6 +49,13 @@
     /// for all marks on a path. If set to false, marks get placed after the
     /// shape they are placed on got transformed.
     transform-shape: true,
+    /// Mark anchor used for placement
+    /// Possible values are:
+    ///   - "tip"
+    ///   - "center"
+    ///   - "base"
+    ///   - "inset"
```
fenjalien, line 57: Is inset supposed to be here?

johannes-wolf:
Can we merge this?

fenjalien (approved):

Issue #463, opened 2024-01-17
Add identifier to mark drawables 


fenjalien:
A new key should be added to drawables that come from marks so they can be identified by other functions so they can be handled differently. This would be useful for such cases as intersections (don't intersect marks), merge-path (don't join marks) and decorations (don't decorate the marks). I'm not sure about the behaviour of a bounding box however.

The name could be as verbose as `is-mark` or as a more general approach `ignore` in the event other elements would produce similar outputs. 

If `ignore` is chosen its difference to hide should be noted, ignored drawables still get drawn to the canvas but won't necessarily get acted upon.

PR #593, opened 2024-05-17
debug: Tag debug shapes

### LINKED ISSUE #575: canvas debug inverts fill area of merge-path
I noticed a very special edge case of using `canvas(debug: true)` together with fills in `merge-path` and some relative coordinates. I did not perform any further testing to minimize the repro code, so it is possible that the bug can be triggered in more general situations.

When activating the canvas debug plots in the following code, the fill areas of the triangle are inverted:

```typst
#import "@preview/cetz:0.2.2"

#cetz.canvas(debug: false, {  // toggle debug here
  import cetz.draw: *
  
  let end-pos = (0.1, 1)
  let (tri-w, tri-h) = (0.5em, 1em)
  merge-path(
    {
      rect((0,0), end-pos)
      line(
        end-pos,
        (rel: (y: -tri-h)),
        (rel: (tri-w, tri-h / 2)),
        close: true,
      )
    },
    stroke: none,
    fill: blue,
  )
})
```

Without debug, expected form:
image
With `debug: true`, unexpected fill area inversion:
image

Maybe I am also abusing the `merge-path` function here since the rect line draw order is not specified in the manual?
My goal was to avoid the fine white line between the rect and the triangle, which happens without `merge-path` in some PDF viewers.

johannes-wolf:
Tag debug paths as such and skip them in `merge-path`.

Debug shapes bring other problems, maybe we should remove the feature or remove documentation about it.

Fixes #575

### DIFF
--- a/src/draw/shapes.typ
+++ b/src/draw/shapes.typ
@@ -1452,7 +1452,7 @@
             }
           }
           for drawable in r.drawables {
-            if drawable.hidden { continue }
+            if drawable.hidden or "debug" in drawable { continue }
             assert.eq(drawable.type, "path")
             segments += drawable.segments
           }
--- a/src/drawable.typ
+++ b/src/drawable.typ
@@ -30,7 +30,7 @@
   }
 }
 
-#let path(close: false, fill: none, stroke: none, segments) = {
+#let path(close: false, fill: none, stroke: none, segments, debug: false) = {
   let segments = segments
   // Handle case where only one segment has been passed
   if type(segments.first()) == str {
@@ -53,7 +53,9 @@
     stroke: stroke,
     hidden: false,
     bounds: true,
-  )
+  ) + if debug {
+    (debug: true)
+  }
 }
 
 #let content(pos, width, height, border, body) = {
--- a/src/process.typ
+++ b/src/process.typ
@@ -44,7 +44,8 @@
         (bounds.low.at(0), bounds.high.at(1), 0)
       )),
       stroke: red,
-      close: true
+      close: true,
+      debug: true,
     ))
   }
 


fenjalien:
This is would also resolve #463 if you used the `ignore` key instead

johannes-wolf:
They are different. Marks must be considered for bbox calculation, debug shapes not. Although, this is not yet implemented in this PR.

fenjalien:
I don't think marks should be considered for intersections and would they be big enough to really count towards a bounding box?

Besides that, if we do have separate keys for debug, marks, hidden and ignore, would there be a way to combine them into one field? Maybe a "type" or "class" key that holds "debug" or "mark", it could even be an array to hold multiple if needed. It could even be a bytefield...
We'll be needing something similar for modifiers anyway.

johannes-wolf:
> I don't think marks should be considered for intersections and would they be big enough to really count towards a bounding box?
> 
> Besides that, if we do have separate keys for debug, marks, hidden and ignore, would there be a way to combine them into one field? Maybe a "type" or "class" key that holds "debug" or "mark", it could even be an array to hold multiple if needed. It could even be a bytefield... We'll be needing something similar for modifiers anyway.

Yes, I think some kind of "class" is good. We could the provide a function for filtering out certain elements.

fenjalien:
If we are still aiming for modifiers in 0.3.0, can this be blocked until modifiers are implemented? Then we can use whatever system works for modifiers.

johannes-wolf:
> If we are still aiming for modifiers in 0.3.0, can this be blocked until modifiers are implemented? Then we can use whatever system works for modifiers.

Yes.

PR #595, opened 2024-05-18
canvas: Add runtime version check for add-ons

johannes-wolf:
This allows other libs checking the cetz version number at runtime for compat.
Code can require a minimum version using the new `assert-version` element (as long as the internal API does not completely change).

Example (`somelib/shapes.typ`):
```
#let my-shape(...) = {
  import cetz.draw: *
  assert-version(version(0,2,2), hint: "somelib")

  rect(...)
}
```

This is a solution to improve the current module/version situation until Typst supports some kind of version negotiation.


johannes-wolf:
This is wanted/needed for the `cetz-plot` split!

fenjalien (approved):
I'm not particularly happy about having the version in `ctx` but I can't think of a better way to do it without the user explicitly having to pass the cetz module (which is ofc worse).

PR #597, opened 2024-05-19
deps: Update dependency versions + CI

johannes-wolf:
Since Typst 0.11.1 is a bugfix release it is ok to use that on the CI, even that cetz required min version is 0.11.0.


requirements.typ
```diff
@@ -1,4 +1,4 @@
 #import "/src/deps.typ"
 #import "@preview/tidy:0.1.0"
-#import "@preview/tidy:0.2.0"
+#import "@preview/tidy:0.3.0"
```
fenjalien, line 3: Why do we import tidy twice? And given the current progress on the docs I don't think we'll need to update tidy.

johannes-wolf: Because some parts of the docs only work with tidy 0.1.0.
As long as we do not have full online docs, we should update the manual.

JamesxX: I think if you're importing it twice, the second `tidy` module shadows the first, no?

PR #598, opened 2024-05-20
licence: Change licence to GPLv3

johannes-wolf:
Reasons: We can include GPL code in plug-ins.


johannes-wolf:
> Are we talking wasm plugins? I don't think that would require relicensing to GPL.

I think it requires it. The library I am using is GPLv3, if we share a compiled binary, we have to be GPL compatible.

I also see no problem with cetz being GPL.

Jollywatt:
What does this mean for fletcher? I don't know much about licenses, but fletcher is currently MIT.

johannes-wolf:
Current cetz is licensed under LGPLv3, so it should not affect consuming packages at all.

AhmedSalih3d:
Why the change from MIT to LGPLv3? What is the intention/thoughts and how does it affect a user, using the library?

Does it mean that any image made in CeTZ must now have all its source code shared ?

johannes-wolf:
Initial idea was GPLv3 to be able to consume GPL-Code. But for some reason people thought that documents using cetz must be open-sourced (which I *think* is not correct). The same goes for packages that consume GPL-Code. As long as they are private, you do not have to open-source them. *BUT* I am not an expert, so this is only how I think it works.

So I've changed to LGPL, which means you can link to cetz even in closed-source packages, and you only have to open-source modified versions of cetz itself.

johannes-wolf:
Gimp is GPL and you do not have to "open source" any graphics made with it: Products made using GPL software must not be GPL. Same with GCC which is also GPL/LGPL.

johannes-wolf:
> > Initial idea was GPLv3 to be able to consume GPL-Code. But for some reason people thought that documents using cetz must be open-sourced (which I _think_ is not correct). The same goes for packages that consume GPL-Code. As long as they are private, you do not have to open-source them. _BUT_ I am not an expert, so this is only how I think it works.
> > So I've changed to LGPL, which means you can link to cetz even in closed-source packages (which are not a thing in Typst, because there are no compiled packages), and you only have to open-source modified versions of cetz itself.
> 
> Straight up gplv3 would mean every package relying on Cetz would also need to be gplv3

Yes, but it would not apply to graphics generated by cetz.

Enivex:
Yes, that's true

PR #602, opened 2024-05-21
line: Add centroid anchor

johannes-wolf:
This adds a `centroid` anchor to closed lines (polygons) and `merge-path`.


JamesxX (approved):
Looks good to me. If you are trying to find a name, it might be "centre of gravity" by the looks of it (but it has been a while since I studied geometry)

src/draw/shapes.typ
```diff
@@ -490,20 +491,25 @@
...
 /// *Root:* `line`
 ///
 /// Supports mark styling.
-/// 
-/// ## Anchors
-/// Supports path anchors.
+///
+/// = Anchors
+///   Supports path anchors.
+///   / centroid: The centroid anchor is calculated for _closed non self-intersecting_ polygons if all vertices share the same z value.
```
fenjalien, line 501: I need to write a styling guide for the new docs but heres a quick run down:
- Use `##` for the anchors heading instead of `=`
- Use a list for anchor names and wrap the name in double `*` so `- **centroid**: `
- wrap in single `*` to make italics not underscores

src/draw/shapes.typ
```diff
@@ -490,20 +491,25 @@
...
+///
+/// - ..pts-style (coordinates, style): Positional two or more coordinates to draw lines between. Accepts style key-value pairs.
+/// - close (bool): If true, the line-strip gets closed to form a polygon
+/// - name (none,string):
```
fenjalien, line 505: Parameters are already defined above

src/draw/shapes.typ
```diff
@@ -1385,6 +1398,12 @@
 ///
 /// Elements hidden via @@hide() are ignored.
 ///
+/// = parameters
+///
+/// = Anchors
+///   / centroid: Centroid of the _closed and non self-intersecting_ shape. Only exists if `close` is true.
+///   Supports path anchors and shapes where all vertices share the same z-value.
+///
```
fenjalien, line 1406: same notes as before, remove parameters heading, and move the rest to the existing anchors subheading below

src/draw/shapes.typ
```diff
@@ -555,10 +561,17 @@
       close: close
     )
 
+    // Find center for simple polygons, might return none
+    let centroid = if close {
+      polygon.simple-centroid(pts)
+    }
```
fenjalien, line 567: could this be done within the named anchor function? it would then be cached by typst when its needed instead of always

src/polygon.typ
```diff
@@ -57,3 +59,48 @@
...
+
+  let (mx, my, mz) = (0, 0, 0)
+  for p in points {
+    let (x, y, z) = p
+    mx += x
+    my += y
+    mz += z
+  }
+  return (mx / 3, my / 3, mz / 3)
+}
```
fenjalien, line 75: This function feels redundent if we have a general method anyway

fenjalien (changes requested):

src/polygon.typ
johannes-wolf: Gives correct points for 3D triangles, the other might not.

fenjalien:
I think you've messed up the git history again

fenjalien (approved):

PR #608, opened 2024-05-25
shapes: Fixes the center anchor for content

### LINKED ISSUE #533: content: Second coordinate sign ignored
When spanning content by passing two coordinates, the signs of the second coordinates values are ignored.

```
content((0,0), (+1,-1), [Test])
content((0,0), (-1,+1), [Test])
```

Both contents are placed at the exact same location, which is incorrect.

johannes-wolf:
Fixes #533.


fenjalien:
changelog. Is the behaviour you noticed/fixed significant to be noted in the docs?

johannes-wolf:
No, it was a bug. The new behaviour is the expected one.

fenjalien (approved):

PR #635, opened 2024-06-27
mark: Respect transform-shape style

johannes-wolf:
This improves & fixes the `mark` function by allowing to pass no secondary argument (defaulting to `0deg`) and respecting the `transform-shape` style.


johannes-wolf:
Will be implemented by postponing mark rendering (by default).

PR #636, opened 2024-06-29
shapes: Circle transform-shape style

johannes-wolf:
Possible implementation of circles that do not have their shape/radius transformed.
I am not sure about this one. It seems useful to have this option, but it is not possible to implement this for rect and other shapes, for instance.

### DIFF
--- a/src/draw/shapes.typ
+++ b/src/draw/shapes.typ
@@ -33,6 +33,7 @@
 /// *Root:* `circle`
 /// == Keys
 ///   #show-parameter-block("radius", ("number", "array"), [A number that defines the size of the circle's radius. Can also be set to a tuple of two numbers to define the radii of an ellipse, the first number is the `x` radius and the second is the `y` radius.], default: 1)
+///   #show-parameter-block("transform-shape", ("boolean"), [If false, only the center of the circle is affected by the current transformation], default: true)
 /// 
 /// = Anchors
 ///   Supports border and path anchors. The "center" anchor is the default.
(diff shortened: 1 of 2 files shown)


fenjalien:
Comparing this to tikz, they get around it by transforming only coordinates by default. They then have a `transform shape` switch which causes the shape of a node to be transformed. 

To also clarify on why this wouldn't work when drawing rectangles, the two coordinates that define the shape would be transformed and the rectangle would be created from those positions. Creating rectangle of a size the user probably wouldn't expect when compared to tikz. I would then suggest adding alternative ways of drawing rectangle. I know processing has `rectMode` which allows you to swap between defining the corners of a rectangle (which is what we do) and a corner/center with a width and height. 

A bigger issue is what happens when drawing custom shapes with a group? The coordinates that make up the lines would be affected and change the shape instead of the position of the whole group...

johannes-wolf:
> Comparing this to tikz, they get around it by transforming only coordinates by default. They then have a `transform shape` switch which causes the shape of a node to be transformed.
> 
> To also clarify on why this wouldn't work when drawing rectangles, the two coordinates that define the shape would be transformed and the rectangle would be created from those positions. Creating rectangle of a size the user probably wouldn't expect when compared to tikz. I would then suggest adding alternative ways of drawing rectangle. I know processing has `rectMode` which allows you to swap between defining the corners of a rectangle (which is what we do) and a corner/center with a width and height.
> 
> A bigger issue is what happens when drawing custom shapes with a group? The coordinates that make up the lines would be affected and change the shape instead of the position of the whole group...

If we want something similar, I guess we could mark non-user supplied coordinates as such and do not transform them (transform them at the root canvas scope only) depending on a switch.

PR #638, opened 2024-07-01
Change docstrings to markdown

fenjalien:
I'm still working on it and I'm pretty happy with it so far. All documenation shall remain within the cetz repository as extended markdown. All docstrings on functions will be converted into a `.mdx` file in `docs/_generated` by a script in the docs repository which can then be imported to the corresponding file. 

I'm sure we'll switch back to Typst once HTML export exists.


src/bezier.typ
```diff
@@ -318,20 +315,15 @@
   }
 }
 
-/// Shorten curve by distance d. This keeps the curvature of the
-/// curve by finding new values along the original curve.
-/// If d is positive the curve gets shortened by moving s
-/// closer to e, if d is negative, e is moved closer to s.
-/// The points s and e are moved along the curve, keeping the
-/// kurves curvature the same (the control points get recalculated).
+/// Shorten curve by distance `d`. This keeps the curvature of the curve by finding new values along the original curve. If `d` is positive the curve gets shortened by moving `s` closer to `e`, if `d` is negative, `e` is moved closer to `s`. The points `s` and `e` are moved along the curve, keeping the curve's curvature the same (the control points get recalculated).
```
johannes-wolf, line 318: Why such super long lines? I prefer a max line length of 110 chars.

fenjalien: Oh I have word wrap turned on so I don't notice how long they get ':) I can change that if you really want.

docs/api/internal/vector.mdx
```diff
@@ -0,0 +1,12 @@
+---
+title: Vector
+---
+
+# {{vector}}
+
+A three element <Type>array</Type> of <Type>float</Type>s.
```
johannes-wolf, line 7: It can be any size.

docs/api/libraries/palette/index.mdx
```diff
@@ -0,0 +1,5 @@
+# Palette
+
+A palette is a function of the form `index => style` that takes an index, that can be any <Type>in?</Type> and returns a canvas style dictionary. If passed the string `"len"` it must return the length of its unique styles. An example use for palette functions is the plot library, which can use palettes to apply different styles per plot.
```
johannes-wolf, line 3: `in?` -> `int`

justfile
```diff
@@ -15,8 +15,5 @@ test *filter:
 update-test *filter:
   typst-test update {{filter}}
 
-manual:
```
johannes-wolf, line 18: Can we add a new shortcut for `doc/` regeneration?

johannes-wolf: … ah, this is done from the docs repo side.

src/bezier.typ
fenjalien: Not fixing as a change in the regex is required for parameter descriptions which makes this difficult...

johannes-wolf (approved):

PR #641, opened 2024-07-03
Allow optionally specifying branch lengths for trees

kollienne:
This treats any pair that begins with a number (2, "a") as a branch-length and child node, allowing arbitrary branch lengths in trees.

I'm open to feedback on this, as it is it probably breaks existing documents where trees are labelled using integers.


johannes-wolf:
Thank you! Can you please add a test case that shows this feature?

PR #656, opened 2024-07-18
Ortho Enhancements

johannes-wolf:
This PR implements:
- Depth ordering inside the `ortho` environment
- Face culling inside the `ortho` environment
- Added the missing export for the `scope` element
- All shapes are drawn CCW

For some reason, all test images changed (non-visibly).

All default CeTZ-shapes are now drawn CCW, to make face-culling work correctly with them; this is a _breaking change_! But they should have been CCW from the start… so better break it now than later.


src/draw/projection.typ
```diff
@@ -18,23 +14,72 @@
...
 
-// Pushes a view- and projection-matrix to transform all `body` elements. The current context transform is not modified.
+#let ortho-projection-matrix = (
+  (1, 0, 0, 0),
+  (0, 1, 0, 0),
+  (0, 0, 0, 0),
+  (0, 0, 0, 1),
+)
+
+// Sort drawables by median or max segment z-value
```
fenjalien, line 24: It doesn't look like it sorts by median?

src/draw/projection.typ
```diff
@@ -18,23 +14,72 @@
...
 // - body (element): Elements
 // - view-matrix (matrix): View matrix
 // - projection-matrix (matrix): Projection matrix
-// - reset-transform (bool): If true, override (and thus ignore)
-//   the current transformation with the new matrices instead
-//   of multiplying them.
-#let _projection(body, view-matrix, projection-matrix, reset-transform: false) = {
+// - reset-transform (bool): Ignore the current transformation matrix
+// - sorted (bool): Sort drawables by maximum distance (front to back)
+// - cull-face (none,string): Enable back-face culling if set to `"cw"` for clockwise
```
fenjalien, line 57: use `str` for the string type

src/polygon.typ
```diff
@@ -0,0 +1,59 @@
...
+  return poly
+}
+
+/// Computes the signed area of a 2D polygon.
+///
+/// The formula used is the following:
+/// $ 1/2 sum_i=0^n-1 x_i*y_i+1 - x_i+1*y_i $
+///
+/// - points (list): List of Vectors of dimension >= 2
+/// -> Signed area
```
fenjalien, line 30: You need to describe what you mean by a polygon, what form should it take, what expectations does the function make about the polygon.

The equation should be katex compatible, *hideous* I know.

Use `array` not `list`

Just use `float` on the return syntax

src/polygon.typ
```diff
@@ -0,0 +1,59 @@
...
+    a += x0 * y1 - x1 * y0
+  }
+  return .5 * a
+}
+
+/// Returns the winding order of a 2D polygon
+/// by using it's signed area.
+///
+/// - point (list): List of polygon points
+/// -> "ccw" (counter clock-wise) or "cw" (clock-wise) or none
```
fenjalien, line 49: `list` to `array`

`-> str,none` put the two return strings in the description

src/polygon.typ
```diff
@@ -0,0 +1,59 @@
+/// Returns a list of polygon points from
+/// a list of segments.
+///
+/// Cubic segments get linearized by sampling.
+///
+/// - segment (list): List of segments
+/// - samples (int): Number of samples
+/// -> List of vectors
```
fenjalien, line 8: - Whats a polygon point
- use `array` instead of `list`
- set the return type as `array` and describe them as an array of vectors in the description.

fenjalien (changes requested):
I need to make a docs guide...

src/draw/projection.typ
johannes-wolf: Oops, forgot to remove that commen.

fenjalien (approved):

RaulDurand:
Any chance to fix #807 ?

PR #658, opened 2024-07-21
Validate Element Names

johannes-wolf:
- Process: Check that element names do not contain '.'
- Canvas: We use `kind` everywhere else as it does not shadow Typst's `type` function + unpacking the array


JamesxX (approved):

fenjalien (commented):
Changelog!

fenjalien (approved):

PR #668, opened 2024-08-07
draw: Function for finding closest point

johannes-wolf:
Adds a new element `closest-point` that computes the closest point between a reference point and one or more elements.


johannes-wolf:
I guess we can add this to 0.3. What do you think @fenjalien?

Issue #47, opened 2023-05-24
Allow setting anchors in the root scope


johannes-wolf:
Allow setting anchors without using a group.
```
#canvas({
  anchor("A", (1,1))
  ...
})
```

Current solution is to use a named group:
```
#canvas({
  group(name: "nodes", {
    anchor("A", (1,1))
  })
})
```

jdbosser:
Is this feature released yet? I tried to create an anchor in the root, and cetz yells at me that it found an anchor outside of a group.

johannes-wolf:
No, this is not implemented. Should I re-open the issue?

jdbosser:
I think so, if the feature is still planned. If you have decided that this is no longer on the roadmap, then I think it should be left closed.

PR #679, opened 2024-08-11
Allow anchors outsides groups

johannes-wolf:
Issue #47.


src/draw/grouping.typ
```diff
@@ -337,17 +337,24 @@
   )
 },)
 
-/// Creates a new anchor for the current group. This element can only be used inside a group otherwise it will panic. The new anchor will be accessible from inside the group by using just the anchor's name as a coordinate.
+/// Creates a new anchor.
```
fenjalien, line 340: We should keep some description of how to use it and what happens in and out of a group

fenjalien (approved):

PR #682, opened 2024-08-13
Release CeTZ v0.3.0

johannes-wolf:
Waiting for the documentation.


typst.toml
```diff
@@ -11,5 +11,5 @@ authors = [
 categories = [ "visualization" ]
 license = "LGPL-3.0-or-later"
 description = "Drawing with Typst made easy, providing an API inspired by TikZ and Processing. Includes modules for plotting, charts and tree layout."
-keywords = [ "draw", "canvas", "plot", "chart", "tree" ]
```
fenjalien, line 14: do you think we'll get away with adding "cetz"

johannes-wolf: Why should we? Keywords are used for discovery. If you do not know cetz, how would you find it using the search?

fenjalien (approved):

PR #688, opened 2024-08-19
Fix flat-brace flipped content anchor

### LINKED ISSUE #687: Possible bug: flat-brace content anchor is misplaced if flip is set to true
As the title says. This becomes evident by looking at the position of the `h` label if one compares:

```typst
decorations.flat-brace((0,0),(10,0), amplitude: 1, debug: true                 )
decorations.flat-brace((2,0),(10,0), amplitude: 1, debug: true, flip: true)
```

In the first case the anchor labelled `h` is where one would expect to find it. In the second case, it's not! Am I missing something or this is a bug?

johannes-wolf:
The whole function needs a rewrite (not for 0.3, though)!

Fixes #687


fenjalien (commented):
Can some tests be added so this doesn't happen again?

PR #708, opened 2024-10-06
Tiny Doc Fixups

johannes-wolf:


fenjalien:
I've made edits to this in #710. We can close this.

PR #709, opened 2024-10-06
Typst 12 Changes

johannes-wolf:
- Updates the test reference images for Typst 12 rc1
- Update the Typst version the CI uses to 12 rc1
- Fixes an issue with typsts `rotate` default behaviour


tests/content/rotate/ref/1.png
fenjalien, line 1: This is a significant difference

tests/group/transform/ref/1.png
fenjalien, line 1: The line thickness has changed? but only slightly?

fenjalien (commented):
I don't think this should be merged into the release pr

tests/group/transform/ref/1.png
johannes-wolf: Looks like some antialiasing change. I think we can ignore that.

tests/content/rotate/ref/1.png
johannes-wolf: Good catch! I've fixed that.

johannes-wolf:
> I don't think this should be merged into the release pr

Rebased onto `master`.

src/draw/shapes.typ
```diff
@@ -921,6 +921,8 @@
         aabb-height,
         border.segments,
         typst-rotate(angle,
+          reflow: true,
+          origin: center + horizon,
```
johannes-wolf, line 925: `center + horizon` is the default, but we should force override it here to prevent show-rules changing cetz' `content` behaviour.

fenjalien (approved):

PR #710, opened 2024-10-07
Fixup docs before release

fenjalien:
Fixes various docs issues, except changing version numbers I suggest we do that in #682


johannes-wolf (approved):
:partying_face: LGTM.

Issue #716, opened 2024-10-10
Inline math content has a zero height


johannes-wolf:
See https://github.com/typst/typst/issues/5180

This is blocking 0.3.0.

PR #717, opened 2024-10-10
Fix Inline Math Content

johannes-wolf:
Using `content` with inline math gives a zero height content element.
Fixes a bug with Typst 0.12.0-rc2 and cetz.

See #716


typst.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "cetz"
 version = "0.3.0"
-compiler = "0.11.0"
+compiler = "0.12.0"
```
fenjalien, line 4: Does this version actually require 0.12.0? The `top-edge` and `bottom-edge` parameters do still exist in 0.11.0

johannes-wolf: Yes, but the real fix for this to work correctly with math content is in Typst 0.12, so I think requiring it is fine.

johannes-wolf:
There is still a problem with text alignment. Do not merge.

fenjalien (approved):

PR #729, opened 2024-10-19
Streamline Angle API Behaviour

### LINKED ISSUE #712: angle results are hard to predict and control
The `angle` API is a bit tricky to use in programmatic drawing, where the angle value is not known statically. One must play with `inner` and `mark` to get the right result depending on the angle value. For example I often need figures for different trigonometric angles:

```typ
#let trig-angle(alpha) = cetz.canvas(length: 2cm, {
  import cetz.draw: *

  let O = (0, 0)
  let P = (1, 0)
  let Q = (alpha, 1)

  line(O, P, mark: (end: ">"))
  line(O, Q, mark: (end: ">"))
  cetz.angle.angle(O, P, Q, mark: (end: ">"), label: [#alpha], label-radius: 150%)
  rect((-1, 0), (1, 0), stroke: none) // for bounding box
})

#for alpha in (60deg, 240deg, -60deg, -240deg) {
  box(trig-angle(alpha))
}
```

image

The first drawing (60deg) is what I expected, but the others all require different fixes:

* For 240deg I need `inner: false`
* For -60deg I must replace `end: ">"` with `start: ">"`
* For -240deg I need both of the above

(Here I could implement the four cases with conditional logic, but in general the angle can be specified by all sorts of coordinates with no explicit angle value.)

What would you think of the following changes to the API:

* make `angle(O, A, B)` always draw from OA to OB (this makes the effect of mark `start` and `end` more predictable)

* allow `inner: auto`, in which case the angle is drawn counterclockwise for positive values, and clockwise for negative values (I think this should be the default, but having this opt-in would be fine)

johannes-wolf:
Angles are now always drawn counter-clock-wise.

Fixes #712


knuesel:
Thanks! With that PR and the following code:

```typ
#let trig-angle(alpha) = cetz.canvas(length: 1.8cm, {
  import cetz.draw: *

  let O = (0, 0)
  let P = (1, 0)
  let Q = (alpha, 1)

  line(O, P, mark: (end: ">"))
  line(O, Q, mark: (end: ">"))
  content((P, -6pt, O), $P$)
  content((Q, -6pt, O), $Q$)

  cetz.angle.angle(O, P, Q, mark: (end: ">"), label: [#alpha], label-radius: 170%)
  rect((-1, 0), (1, 0), stroke: none) // for bounding box
})

#for alpha in (60deg, 240deg, -60deg, -240deg) {
  box(trig-angle(alpha))
}
```

I get this:

image

and if I add `inner: false` I get:

image

so the functionality is perfect for my use case.

I have two concerns though:

1. The keyword name `inner` doesn't make sense now
2. Other users might miss the previous functionality (having an easy way to choose between "short" and "long" path for the angle)

What about replacing `inner` with a `direction` keyword, with the following possible values:

* `"clockwise"`
* `"counter-clockwise"`
* `"near"`
* `"far"`

I think this way it would be easy to achieve the desired result in all common use cases. What do you think?

johannes-wolf:
I've changed `inner:` to `direction:` with the value `"cw"` and `"ccw"`.

knuesel:
Excellent! I think that new API is great and something like "near"/"far" could be added later if there's demand for it.

PR #747, opened 2024-11-11
Fix docs for `draw-edge` function in `tree` library

Lipen:
Fix the signature of `draw-edge` function argument in the `tree` library to correctly specify that it actually accepts 4 arguments.


johannes-wolf (approved):

johannes-wolf:
Thank you.

PR #756, opened 2024-11-24
decoration: Allow dynamic amplitudes

johannes-wolf:
Allows passing a function of the form `ratio -> float` or an array of floats as `amplitude:` to path decorations. 

Adds a `square` square-wave decoration.


src/lib/decorations/path.typ
```diff
@@ -59,6 +63,16 @@
   factor: 150%,
 )
 
+#let resolve-amplitude(amplitude, segment, num-segments) = {
+  return if type(amplitude) == function {
+    (amplitude)(segment / num-segments * 100%)
```
Jollywatt, line 68: I think it's better if this is `segment / (num-segments - 1) * 100%`, as then the argument will range from `0%` to `100%` inclusive for the first and last segments. Then the output will appear symmetric for symmetric amplitude functions like `t => calc.min(float(t), 1 - float(t))`.

Jollywatt:
One more nitpick:
```
  //   ╭ ma ╮        ▲
  //   │    │        │ Up
  // ..a....m....b.. '
  //        │    │
  //        ╰ mb ╯
```
Since the wave points `ma` and `mb` are at different positions along the curve, you could evaluate the amplitude at both points independently.
This leads to slightly nicer results such as:
Screenshot 2024-11-24 at 18 44 42
compared to what you currently get:
Screenshot 2024-11-24 at 18 45 04

Both of these are made with `amplitude: t => calc.min(float(t), 1 - float(t))`.
```

For the first option,
```typ

```
are replaced with
```
let fn(i, a, b, norm) = {
    let ab = vector.sub(b, a)
    let up = vector.scale(norm, resolve-amplitude(style.amplitude, i + 0.25, num-segments) / +2)
    let down = vector.scale(norm, resolve-amplitude(style.amplitude, i + 0.75, num-segments) / -2)

File: src/lib/decorations/path.typ [hunk 2]
```diff
@@ -59,6 +63,16 @@
   factor: 150%,
 )
 
+#let resolve-amplitude(amplitude, segment, num-segments) = {
+  return if type(amplitude) == function {
+    (amplitude)(segment / (num-segments - 1) * 100%)
```
Jollywatt, line 68: This now needs to be reverted back to `(amplitude)(segment / num-segments * 100%)`, since `segment` now ranges from `0.25` to `num-segments + 0.75` 😅

You can see what I mean with the test case
```typ
wave(line((0,3), (4,3)), amplitude: t => { calc.sin(t/100%*calc.pi) })
```

johannes-wolf: Yeah, I am playing around with it. With the .25 addition, you won't get `0%` in the callback, which is a bit unexpected, I think.

src/lib/decorations/path.typ
```diff
@@ -59,6 +63,16 @@
...
+  } else if type(amplitude) == array {
+    amplitude.at(int(calc.max(0, calc.round(segment / (num-segments - 1) * (amplitude.len() - 1)))), default: 0)
```
Jollywatt, line 70: Same with this one, looking at test case:
```typc
wave(line((0,4), (4,4)), amplitude: (1,0,1))
```

src/lib/decorations/path.typ
Jollywatt: I guess the `segment` variable is no longer an index, but rather the coordinate along the curve `0 <= t <= 1`

src/lib/decorations/path.typ
johannes-wolf: Changed it.

src/lib/decorations/path.typ
```diff
@@ -59,6 +63,24 @@
...
+  /// Midpoint factor
+  factor: 50%,
+)
+
+#let resolve-amplitude(amplitude, segment, num-segments) = {
+  segment = calc.max(0, calc.min(segment, num-segments))
+  return if type(amplitude) == function {
+    (amplitude)(segment / num-segments * 100%)
+  } else if type(amplitude) == array {
+    amplitude.at(calc.rem(int(segment), amplitude.len()), default: 0)
```
Jollywatt, line 78: Have you considered `amplitude.at(calc.rem(int(2*segment), amplitude.len()), default: 0)`? That gives the user control over individual peaks and troughs. With `int(segment)`, you can only control them in full-wavelength pairs.

Jollywatt (commented):
I just had one possible suggestion:

> Have you considered `amplitude.at(calc.rem(int(2*segment), amplitude.len()), default: 0)`? That gives the user control over individual peaks and troughs. With `int(segment)`, you can only control them in full-wavelength pairs.

johannes-wolf:
> I just had one possible suggestion:
> 
> > Have you considered `amplitude.at(calc.rem(int(2*segment), amplitude.len()), default: 0)`? That gives the user control over individual peaks and troughs. With `int(segment)`, you can only control them in full-wavelength pairs.

Will look into that.

Jollywatt (approved):
Perfect:)
I think `length` amplitudes should also be allowed, but I'm happy to make that a separate PR. It's just a matter of adding `util.resolve-number(ctx, amplitude)` and adding some tests.

PR #761, opened 2024-11-27
content: Clip content size

### LINKED ISSUE #748: Negative padding causes crash
image

johannes-wolf:
fixes #748

Negative padding can result in offset content, but this PRs goal is to prevent the compiler crash.


jamesrswift (approved):
Minimal change and added test - looks good to me!

PR #767, opened 2024-12-06
decoration: Allow specifying amplitudes as lengths

Jollywatt:
This allows the amplitude parameter in path decorations to be given as a length, for example:
```typc
fn(line((0,0), (4,0)), amplitude: 0.25)
fn(line((0,1), (4,1)), amplitude: 2.5mm)
fn(line((0,2), (4,2)), amplitude: t => 1em*calc.sin(float(t)*calc.pi))
fn(line((0,3), (4,3)), amplitude: (5mm, 0, 2mm, 0))
```
for a path decoration function `fn`.


Jollywatt:
This is a follow up from #756.

johannes-wolf (approved):
LGTM.

PR #777, opened 2024-12-28
shapes: Add polygon shape

### LINKED ISSUE #776: Drawing Polygons
I am submitting this request as I have a use case that is somewhat tedious to deal with myself.

Currently, Cetz can draw circles and rectangles. Unless I am mistaken, there is no built-in way to draw an arbitrary polygon (with some radius r).

Currently, in my own project, I am dealing with hexagonal grids, so my code has something like
```typ
// Some initial rotation
let angle = 360deg / 6 + rotation

let verts = ()

// Get the n vertices of the polygon
while angle < 360deg + to-angle(rotation) {
  let vert = (size * calc.cos(angle), size * calc.sin(angle))
  verts.push(vert)

  angle = angle + 180deg / 3
}
```
Then I use this with `draw.line` with the close arg.

Because my project is somewhat complex, I need to do a lot of manual linear algebra since I cannot use intersections or because I have to do my own anchor shenanigans.

Having something built in to provide all those existing features would be a very big help.

To illustrate the difficulties of my use case, here are two images from my project (with and without my own "debug" feature):

The main problem I faced was the difficulty of intersections. For some reason, I could not draw a line between two hexagons. The docs are rather poor on this topic and I have not worked it out. Thus, to ensure the gradient has the correct midpoint, I had to manually compute the start and end points of my lines (so that the lines drew from vertex to vertex and not from centre to centre -- intersections would resolve this). Thus, there was the difficulty of actually using basic draw features aptly in a dynamic setting, along with additional complications involving adequate styling that required some fighting with both the library and typst to make work correct.

Having a function `draw.polygon(sides, radius)` would effectively solve _most_ problems (except the ones that are typst issues). If possible, I would very much love this addition.

Thanks. Sorry if I worded anything strangely.

Edit:

I forgot to mention the pain anchors are for my debug icons/info. But I figure that makes sense how it is a bit of a chore.

johannes-wolf:
Fixes #776.


janosh:
thanks for adding polygons! very useful, excited to try this in the next release.

PR #814, opened 2025-02-16
canvas: Use new curve API

johannes-wolf:


typst.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "cetz"
 version = "0.3.2"
-compiler = "0.12.0"
+compiler = "0.13.0-rc1"
```
epsilonhalbe, line 4: I think version 0.13.0 has been released so the `-rc1` can be dropped

.github/workflows/ci.yml
```diff
@@ -21,12 +21,13 @@ jobs:
...
       - name: Install tytanic
-        uses: baptiste0928/cargo-install@v2.2.0
+        uses: baptiste0928/cargo-install@v3
         with:
           crate: tytanic
+          git: https://github.com/tingerrr/tytanic.git
       - uses: typst-community/setup-typst@v3
         with:
-          typst-version: '0.12.0'
+          typst-version: '0.13.0-rc1'
```
epsilonhalbe, line 30: i think 0.13.0 is released so the `-rc1` can be dropped.

PR #816, opened 2025-02-17
perf: Some performance improvements

johannes-wolf:
Gives ~2 seconds when running tests.

### DIFF
--- a/src/canvas.typ
+++ b/src/canvas.typ
@@ -46,7 +46,7 @@
     transform:
       ((1, 0,-.5, 0),
        (0,-1,+.5, 0),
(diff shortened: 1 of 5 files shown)


src/canvas.typ
```diff
@@ -46,7 +46,7 @@
     transform:
       ((1, 0,-.5, 0),
        (0,-1,+.5, 0),
-       (0, 0, .0, 0),
+       (0, 0,  1, 0),
```
johannes-wolf, line 49: Idk why there was a zero. And the `inverse` function should normally fail to compute an inverse of that matrix.

PR #844, opened 2025-03-20
Add fast path for rect without radius

rikhuijzer:
This PR suggests to add a fast path for rectangles without a radius. When benchmarking with 100 and 500 points, it is about 5% faster on my system. My test document was

```typ
#import "/src/lib.typ": *

#let pts = array.range(500).map(x => (x, x))

#canvas({
  import draw: *

  for pt in pts {
    rect(pt, (rel: (0.5, 0.5), to: pt), fill: blue, stroke: none)
  }
})
```
and my benchmark command was
```sh
$ hyperfine --warmup=1 --runs=10 "typst c --ppi=10 --format=png tmp.typ"
```
which generates a 106 KB 1986 x 1986 image.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 docs/_generated
+tmp*
--- a/src/draw/shapes.typ
+++ b/src/draw/shapes.typ
@@ -1169,7 +1169,26 @@
       let (north-west: nw, north-east: ne,
            south-west: sw, south-east: se) = util.as-corner-radius-dict(ctx, style.radius, size)
 
-      let drawables = {
+      let no-radius = style.radius == none or style.radius == 0
+      let drawables = if no-radius {
+        let segments = ()
+        // Four corner points for a non-rounded rectangle:
+        //
+        //   p0-----p1
+        //   |       |
+        //   |       |
+        //   p3-----p2
+        //
+        let p0 = (x1, y2, z1)
+        let p1 = (x2, y2, z1)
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1 +1,2 @@
-docs/_generated
\ No newline at end of file
+docs/_generated
+tmp*
```
rikhuijzer, line 2: I was using this extra `tmp*` to put my test files in the root of the project with names such as `tmp.typ` and `tmp.png`. I hope it's okay to add this one to the `.gitignore`.

rikhuijzer:
I see the tests fail. I'll investigate.

rikhuijzer:
Locally all tests (including `content`) pass. I'm not sure what to do now

rikhuijzer:
It looks like the test failure is a fluke, locally the failing `content` test passes. Instead, locally `group-transform` fails. However, if I load an image with the old code and the new code into https://www.diffchecker.com/image-compare/, then I see no difference:

afbeelding

The binary files are different though according to `diff`. But the "difference" view locally and on diffchecker.com are completely black images.

johannes-wolf:
Looks good! Just commit the changed images :)

rikhuijzer:
Turns out the tests were right! There were some differences. Fix now in d9c264a. It took me a while to get it but I start to like this testing infrastructure. It's fast and seems quite reliable

johannes-wolf (approved):

PR #845, opened 2025-03-21
Extract `dim-extrema` from `cubic-extrema`

rikhuijzer:
It's quite a silly change, but I'm getting consistently 3%-4% better performance with

```typst
#import "/src/lib.typ": *

#let pts = array.range(500).map(x => (x, x))

#canvas({
  import draw: *

  for pt in pts {
    circle(pt, fill: blue, stroke: none)
  }
})
```
and
```sh
$ hyperfine --warmup=1 --runs=10 "typst c --ppi=10 --format=png tmp.typ"
```

Are you in favor of rewrites like this? Or should I see whether this can be fixed in the Typst compiler? 

I personally would lean in favor of biting the bullet and rewriting these kinds of code because it sort of makes sense that defining a function at the module level once is faster than constantly defining a function as an inner function. Also, readability-wise having the function on the top level is easier to read in my opinion. But I don't have a very strong opinion on this.

### DIFF
--- a/src/bezier.typ
+++ b/src/bezier.typ
@@ -386,6 +386,37 @@
   }
 }
 
+// Compute roots of a single dimension (x, y, z) of the
+// curve by using the abc formula for finding roots of
+// the curves first derivative.
+#let _dim-extrema(a, b, c1, c2) = {
+  let f0 = calc.round(3*(c1 - a), digits: 8)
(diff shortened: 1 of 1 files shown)


johannes-wolf:
I am fine with this, but I think the Typst compiler does something wrong here, as it should not be slower, even if values were captured.

johannes-wolf (approved):

PR #846, opened 2025-03-21
Calculate `cubic-extrema` inside WebAssembly

rikhuijzer:
While working on #845 and #844, I noticed that the Typst language is missing some features and as seen in #845 the compiler also is missing some features. And don't get me wrong. I think the Typst language is **insanely** good given how new it is and all, but of course it cannot compete with the years of engineering that went into a language like Rust. The primary thing that I am missing in the Typst language is types. Especially in CeTZ where a variable can be a `f64`, `Vec<f64>` (point), or `Vec<Vec<f64>>` (points). 

So that got me thinking, can we run the core CeTZ math inside Rust via the wasm plugin system? This PR is an experiment that runs the `cubic-extrema` via wasm. I'm getting some errors now during testing. But I expect those are mostly caused by differences in floating point math. From what I can see, the differences in the generated images are mostly minor. It seems this experiment actually works.

The speedup is about 10%-20%.

For moving variables into the WebAssembly runtime, I used cbor. (de)serialization is necessary because by default we just pass bytes buffers around. I chose cbor instead of json because cbor should be faster. Even with this serialization and deserialization it seems like wasm is quicker, so I guess writing in Rust is really helping in speeding things up. The more you run inside wasm, the smaller the (de)serialization overhead will be.

@johannes-wolf is this a path that is interesting for CeTZ, you think? We have to figure out a way to distribute the wasm and make it easy to work with this during development, but I think it's possible. There are already multiple Typst packages that use wasm.


rikhuijzer:
To verify that my claim about that the failing tests are probably due to minor differences, here is my benchmark image from #845 with 20 points when created via wasm:

```sh
$ typst c --ppi=100 --format=png tmp.typ
```

via wasm

and this when switching back to the old logic:

without wasm

Also, the `element-anchors` and `content` test for example respectively report only 2 and 1 deviations.

rikhuijzer:
> Perfect! I also wanted to try this (but with C++, because I have no Rust experience).

Thanks for being open to this kind of crazy idea. Nice that you have C++ experience, then Rust should be quite straightforward.

> I wonder if this is also faster in the browser, as I remember reading that Typsts WASM plug-ins are slow in the web environment.

I just ran it in the browser and extracted the timing info via "Help -> Debug -> Restart and start profiling". From that, I got the following results.

When plotting only 50 points, the loop in `/src/process.typ#76` is faster with 67 ms versus 77 ms, **but** the total runtime is slower since a lot of time is spent in `eval` 117 ms vs 240 ms. My guess is that this is loading of the wasm library (but this might be incorrect as I discuss in the end? I'm not sure about the eval).

However! For more points the wasm version is faster again. For this script:
```typ
#import "/src/lib.typ": *

#set page(width: 7in, height: 7in)

#let pts = array.range(500).map(x => (x/10, x/10))

#canvas({
  import draw: *

  for pt in pts {
    circle(pt, fill: blue, stroke: none)
  }
})
```
the non-wasm version takes 1 seconds 200 ms for the compile block, while the wasm version takes 890 ms (-25%). 

Just to be sure, I ran another benchmark but now with 100 points. Then, eval takes 23 ms and the full compile 219 ms without wasm. With wasm, the eval takes 25 ms and the full compile takes 187 ms (-15%).

image

And another run on 2000 points, now the full compile took 4s 190ms, of which 371 eval time, without wasm and 3s 570ms, of which 238 ms eval, with wasm (-15%).

It seems sometimes `eval` messes the timing information up. I'm not sure whether there is just a wide range of possible eval values or maybe that eval is part something else? I'm not sure. Overall the compilation seems consistently faster with wasm.

I'll do some more tests in the next few days and update the `justfile` and other things in this PR to streamline the development workflow with wasm.

rikhuijzer:
Unlike what I said above, the tests now pass even when running via wasm. I accidentally was using `f32` at the Rust side because I thought Typst uses 32-bit floats, but actually Typst uses 64-bit floats (https://typst.app/docs/reference/foundations/float/).

I re-ran the hyperfine benchmark and basically got the same results as before.

For development, I have now added the `cargo build` to the `justfile`. `just install` and `just test` will now automatically build the wasm file before running.

The wasm build requires Rust to be installed on the system of the developer. For most Rust developers this should already be available. Otherwise, the required developer tools can be installed via

```sh
# Source: https://rustup.rs/.
$ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh

# Install latest stable Rust version.
$ rustup update stable

# Install the wasm compilation target.
$ rustup target add wasm32-unknown-unknown
```

For myself, to get automatic rebuilds when I change a file, I was using:
```sh
$ find src/**/*.typ cetz-core/**/*.rs | entr -s "just test"
```
This automatically runs the tests each time I update a Typst or Rust file.

johannes-wolf:
Looks good! I will test it on my machine and then merge it.

In the future I would like to port more functions over to Rust.
After #843, adding support for bounds calculation & transformation of whole paths in Rust might improve perf. further.

We should look for some vector/matrix lib for rust to not having to implement all the basic functions again.

johannes-wolf (approved):
LGTM

rikhuijzer:
> We should look for some vector/matrix lib for rust to not having to implement all the basic functions again.

[ndarray](https://github.com/rust-ndarray/ndarray) could be a good option. It has some basic vector operations like `dot` and transpose, and seems well maintained. Also, it runs well in the wasm32 target (I've tested) and can even be used in `[no_std]` environments (`[no_std]` could be useful for CeTZ too at some point because it will reduce the binary size).

Alternatively [nalgebra](https://nalgebra.org) could be a good option too. They have some [docs for wasm](https://nalgebra.org/docs/user_guide/wasm_and_embedded_targets) even.

PR #849, opened 2025-03-24
Use dictionary for segments

rikhuijzer:
I couldn't move the `path-util.bounds` calculation into WebAssembly because a segment had type `Vec<String, Points>`. So I turned segments into a dictionary `(kind: kind, points: pts)`. Now it worked.

Although the dictionary is arguably a bit easier to understand, ~~my benchmark runs a bit slower on this PR than current `master` branch (+2%). (Overall, the benchmark still runs faster than before #846.)~~ edit: this PR is a bit faster compared to `master`. I thought it was slower but that was due to an assertion I added (fixed in 3dd0925).

It's up to you @johannes-wolf to decide whether you are interested in this change. I guess it makes sense to refactor the segment type, but I can also understand it if you prefer the old code.


johannes-wolf:
It conflicts with my changes here: #843 :( I wanted to change the paths to a collection of sub-paths with an origin and one or more drawing commands ("l" and "c").

rikhuijzer:
> It conflicts with my changes here: #843 :( I wanted to change the paths to a collection of sub-paths with an origin and one or more drawing commands ("l" and "c").

No problem! Feel free to finish #843 first and I'll update this PR after that :)

johannes-wolf:
#843 finally got merged :)

PR #850, opened 2025-03-24
Run `aabb` in wasm

rikhuijzer:
This is 30% faster. Plotting 1000 circles now takes less than 1 s on my system 🚀 (down from 1.6 s).

This is mostly due to removing this line though:
```typst
assert(type(pt) == array and pt.len() == 3, message: repr(init) + repr(pts))
```
This check is now still done inside wasm. If the `pt` is not an array the program will error when trying to parse the incoming cbor data.

Also, Rust found a small bug, if you can call it that. Usually a `pt` is a `Vec<f64>`, but sometimes inside CeTZ a `0` (integer) was introduced instead of a float.

This PR depends on #848. I'll leave this one on draft until the other PR is merged.


rikhuijzer:
> Can we make the cbor decoding more relaxed about `int` vs `float`?

Maybe it would be an option for cetz to promote the integers to floats after entering cetz? Both integers and floats are [stored in 64-bit types](https://typst.app/docs/reference/foundations/int/), so performance-wise it shouldn't matter I think

PR #852, opened 2025-03-28
WIP: Rust Experiments

johannes-wolf:
Some tests.


johannes-wolf:
@rikhuijzer: Sume Rust experiments for modifying complete paths on the Rust side.
One problem I could not solve is `Point`s with integers instead of floats. Serde fails if the "wrong" type is in the cbor and there seems to be no option to accept ints where floats are expected.

cetz-core/src/lib.rs
```diff
@@ -103,3 +114,85 @@ pub fn cubic_extrema_func(input: &[u8]) -> Vec<u8> {
         }
     }
 }
+
+#[derive(Serialize,Deserialize)]
+#[serde(untagged)]
+enum Segment {
+  Line(String /* "l" */, Point),
```
johannes-wolf, line 121: This is ugly.

cetz-core/src/lib.rs
```diff
@@ -1,11 +1,14 @@
...
+use serde::Serialize;
 use serde::Deserialize;
+use serde_tuple::*;
+use arrayvec::ArrayVec;
 use wasm_minimal_protocol::*;
 
 initiate_protocol!();
 
-type Point = Vec<f64>;
+type Point = ArrayVec<f64, 4>;
```
rikhuijzer, line 11: Yes this is great. I was also thinking about something like this

rikhuijzer:
Experimenting. I like it

> Serde fails if the "wrong" type is in the cbor and there seems to be no option to accept ints where floats are expected.

Yes I also ran into this a few times. This is Rust’s blessing and curse I think. I was fixing it by trying to be more strict at the Typst side. I fixed one int case in I believe #849 (on phone now so can’t check) which didn’t seem too bad. Naja I’m also open to other ideas these were just my 2 cents

rikhuijzer:
I think static typing is a massive improvement in programming in general, but people are allowed to disagree with me on this one

PR #873, opened 2025-05-15
path-util: be more forgiving of floating point errors in assertion

Jollywatt:
I observed panics occurring when setting `cetz.decorations.wave(segment-length: ..)` due to floating point error. (This led to some unpredictable crashes in fletcher.)

For example:
```typ
#test-case({
  import draw: *
  wave(line((0,0), (3,0)), segment-length: 0.333)
})
```
causes `assertion failed: t is expected to be between 0 and the length of the path (3), got: 3.0000000000000004`.

We could track down the origin of this the floating point error and try to fix it, or we could just be a bit more forgiving and not throw assertion errors if `t` is only slightly over the maximum path length (this is what I did).


johannes-wolf:
With [#path-mod](https://github.com/cetz-package/cetz/pull/843) I completely removed that assertion.

Jollywatt:
> With [#path-mod](https://github.com/cetz-package/cetz/pull/843) I completely removed that assertion.

Oh great, I'll abandon this PR then! :)

PR #877, opened 2025-06-04
Use `handle_cbor` for `aabb_func`

rikhuijzer:
A small simplification for `aabb_func`. 

Also I've introduced a helper function inside Typst that might be cleaner? Let me know what you think @johannes-wolf. If you don't like it I can turn it back.


johannes-wolf (approved):
Looks good!

PR #879, opened 2025-06-05
Run `mul4x4-vec3` in wasm

rikhuijzer:
As a continuation of https://github.com/cetz-package/cetz/pull/850, this is an experiment which probably will not be merged and that's fine.

I noticed that when running a [benchmark](https://github.com/cetz-package/cetz/pull/850#issuecomment-2936119941) that a lot of time is spent inside `mul4x4-vec3`. So I tried to move that in Wasm to see whether it's faster. Unfortunately, it isn't.

It does pass the tests, but it is slower. 550 ms versus 500 ms (+10%). My guess is that converting (de)serializing vectors takes relatively too much time. I've tried to call wasm directly from inside `apply-transform` inside `drawable.typ`, but that also didn't speed things up. The bottom line is that (de)serializing the `segments` apparently takes too much time. I could try to go even higher and move even more in wasm, but it seems to me that the PR becomes a bit too big then.

In response to [a comment](https://github.com/cetz-package/cetz/pull/850#pullrequestreview-2893970888):

> Can we make the cbor decoding more relaxed about `int` vs `float`?

After working on the code for this PR a few hours, I do think promoting to float and using that internally should be an okay solution. CeTZ anyway promotes integers to floats all the time since it happens anytime basic arithmetic happens between int and float (e.g., 1 + 1.1 = 2.1 and 1 / 2 = 0.5).

### DIFF
--- a/cetz-core/src/lib.rs
+++ b/cetz-core/src/lib.rs
@@ -170,13 +170,13 @@ struct Mul4x4Vec3Args {
     w: f64,
 }
 
-fn mul4x4_vec3(mat: Vec<Point>, vec: Point, w: f64) -> Result<Point, String> {
+fn mul4x4_vec3(mat: &[Point], vec: &Point, w: f64) -> Result<Point, String> {
     assert!(vec.len() <= 4);
 
     let x = vec[0];
     let y = vec[1];
-    let z = vec.get(2).unwrap_or(&0.0);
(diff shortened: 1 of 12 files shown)


johannes-wolf:
I've copied some of the float promotion to #881. I think it should be enough to make sure
the transformation matrix uses floats in all components.

PR #885, opened 2025-06-13
shapes: Add bbox primitive (rect as aabb)

halbuer:
Drawing a border around several elements is a common task when creating scientific drawings to indicate logical connection. The bbox shape is a wrapper for the existing rect primitive that calculates an axis aligned bounding box for a given collection of coordinates or elements. If a shape is given, all of its anchors are taken into account for aabb calculation.

The padding attribute can be used to specify inner spacing of the bounding box. All other styling attributes are forwarded to the rect.


johannes-wolf (commented):
Thank you for your PR. I like the functionality but I am not yet sure about the name as `bbox` could also be a function that returns the bounding box via callback. What do you think about `rect-around`?

halbuer:
Thx for your feedback! I agree, that the term `bbox` could be confusing and changed it to the proposed `rect-around`.

src/draw/shapes.typ
```diff
@@ -1672,3 +1672,60 @@
...
+      }
+    }
+
+    let style = styles.resolve(ctx.style, merge: style, root: "rect", base: (padding: none))
+
+    let (ctx, ..vecs) = coordinate.resolve(ctx, ..more_points)
+    let bounds = aabb.aabb(vecs)
+    // Resolve padding and convert to canvas units
+    let padding = util.as-padding-dict(style.padding)
+    // Swap top and bottom padding for reasons
```
johannes-wolf, line 1718: I think `(padding.top, padding.bottom) = (padding.bottom, padding.top)` looks nicer
for swapping two values.

johannes-wolf (changes requested):
Can you add some tests and rebase on master please?

If running/adding tests does not work, I can also add some post-merge.

For the documentation generation you have to add a file under
`docs/api/draw-functions/shapes/rect-around.mdx` with:
```mdx
import RectAround from "@site/cetz/docs/_generated/draw/shapes/rect-around.mdx";            
                                                                               
# rect-around                                                                         
                                                                               
<RectAround />
```

halbuer:
> Can you add some tests and rebase on master please?
> 
> If running/adding tests does not work, I can also add some post-merge.
> 
> For the documentation generation you have to add a file under `docs/api/draw-functions/shapes/rect-around.mdx` with:
> 
> ```mdx
> import RectAround from "@site/cetz/docs/_generated/draw/shapes/rect-around.mdx";            
>                                                                                
> # rect-around                                                                         
>                                                                                
> <RectAround />
> ```

I added two test cases that should cover most of the features. I also added the doc file as provided but unfortunately the doc generation does not work on my machine, so not tested.

PR #886, opened 2025-06-13
feat: Add nstar function

### LINKED ISSUE #847: N-pointed star
would it make sense to add $N$-pointed stars to the cetz shape repertoire? useful for [diagrams like this one](https://github.com/janosh/diagrams/commit/392835e2927b79af05998c17c03aad5a241e3e9e)

Image

the main arguments besides number of points would probably be inner and outer radius, starting angle and whether or not the lines connecting inner-radius points should be drawn like here:

Image

AndPuQing:
### Summary

This pull request introduces a new `#nstar` function for drawing n-pointed stars, a commonly requested geometric primitive.

### Schläfli Symbols

This implementation defines stars by their configurable inner and outer radii, which is standard and highly flexible for computer graphics. However, it's fascinating to consider this in the context of pure geometry, particularly [**Schläfli symbols (`{p/q}`)**](https://en.wikipedia.org/wiki/Star_polygon) which describe "true" regular star polygons.

- A true star polygon `{p/q}` (e.g., a pentagram `{5/2}`) is a single, non-intersecting path formed by connecting every `q`-th point of `p` points on a circle. This is only possible if `p` and `q` are coprime.
- Under this definition, shapes like the hexagram (Star of David) are not true star polygons but **compounds** of two triangles (derived from `{6/2}`). Similarly, a "square star" is a compound of two line segments.
- In a true `{p/q}` star, the ratio of the inner to outer radius is a fixed mathematical constant, not a free parameter.

Perhaps we could further design a `#regular-shape` function, along with a `#schlafli-shape(p, q)` function that calculates the correct radius ratio and invokes `regular-shape` to render these geometrically "perfect" shapes. This approach could support various shapes such as `triangle`, `square`, `pentagon`, and `pentagram`. Additionally, the existing `polygon` and `nstar` implementations could be migrated to this unified `regular-shape` interface.  

> This is not the final PR, any comments are welcome

**Closes #847**

### DIFF
--- a/src/draw/shapes.typ
+++ b/src/draw/shapes.typ
@@ -694,8 +694,7 @@
 ///
 /// ## Styling
 /// Root: nstar
-/// - inner-radius (number) = 0.5: The radius of the circle on which the inner points lie.
-/// - outer-radius (number) = 1: The radius of the circle on which the outer points lie.
+/// - radius (inner, outer) = : The radius of the star's inner and outer points.
 /// - show-inner (bool) = false: If true, also draws the inner polygon connecting the star's inner points.
 /// - fill (color, gradient): The fill color for the star.
 /// - stroke (color, thickness, ...): The stroke for the star and the inner polygon.
@@ -710,7 +709,7 @@
     let style = styles.resolve(ctx.style, merge: style, root: "n-star")
 
     let (ctx, origin) = coordinate.resolve(ctx, origin)
-    let (inner_radius, outer_radius) = if type(style.radius) == array {
+    let (inner-radius, outer-radius) = if type(style.radius) == array {
       style.radius.map(util.resolve-number.with(ctx))
     } else {
       (style.radius, style.radius).map(util.resolve-number.with(ctx))
@@ -721,14 +720,14 @@
     let point_pairs = range(0, sides).map(i => {
       let inner_angle = angle + i * angle_step
       let inner_point = vector.add(origin, (
-        calc.cos(inner_angle) * inner_radius,
-        calc.sin(inner_angle) * inner_radius,
+        calc.cos(inner_angle) * inner-radius,
+        calc.sin(inner_angle) * outer-radius,
         0
       ))
       let outer_angle = inner_angle + angle_step / 2
       let outer_point = vector.add(origin, (
-        calc.cos(outer_angle) * outer_radius,
-        calc.sin(outer_angle) * outer_radius,
+        calc.cos(outer_angle) * outer-radius,
+        calc.sin(outer_angle) * outer-radius,
         0
       ))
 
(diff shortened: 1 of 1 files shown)


johannes-wolf:
Thank you for your PR! I am not sure about the name `nstar`, would you mind changing it to `n-star`, as it looks more “typst like”?

johannes-wolf (changes requested):
One small thing: add a `ref` directory next to your test file and run `just update-tests` (or `tt update`) to generate the missing reference image.
LGTM otherwise.

AndPuQing:
> One small thing: add a `ref` directory next to your test file and run `just update-tests` (or `tt update`) to generate the missing reference image. LGTM otherwise.
> 
> If you cannot run the tests, let me know, so I can add a ref-image post merge.

lol, I'm still confused as to why `tt`'s output is not traced by git

johannes-wolf:
> > One small thing: add a `ref` directory next to your test file and run `just update-tests` (or `tt update`) to generate the missing reference image. LGTM otherwise.
> > If you cannot run the tests, let me know, so I can add a ref-image post merge.
> 
> lol, I'm still confused as to why `tt`'s output is not traced by git

It is, but `tt` handles `test.typ` files as “compile only” if no `ref` dir is present and such generates no images for the test.

src/draw/shapes.typ
```diff
@@ -674,6 +674,125 @@
...
+///
+/// - origin (coordinate): Coordinate to draw the star's center at.
+/// - sides (int): Number of points of the star (>= 3).
+/// - angle (angle) = 0deg: Angle to rotate the star around its origin.
+/// - name (none, str): An optional name to identify the shape.
+///
+/// ## Styling
+/// Root: nstar
+/// - inner-radius (number) = 0.5: The radius of the circle on which the inner points lie.
+/// - outer-radius (number) = 1: The radius of the circle on which the outer points lie.
```
johannes-wolf, line 698: Radius seems to be a single array `(inner,outer)` but the docstring specifies two distinct arguments.

src/draw/shapes.typ
```diff
@@ -674,6 +674,125 @@
...
+    message: "Invalid number of sides: " + repr(sides))
+
+  let style = style.named()
+  return (ctx => {
+    let anchors = ()
+
+    let style = styles.resolve(ctx.style, merge: style, root: "n-star")
+
+    let (ctx, origin) = coordinate.resolve(ctx, origin)
+    let (inner_radius, outer_radius) = if type(style.radius) == array {
```
johannes-wolf, line 713: Nitpick: use a dash instead of an underscore.

johannes-wolf (changes requested):

johannes-wolf (approved):
LGTM.

PR #891, opened 2025-06-20
Measure Full Tree Node

johannes-wolf:
@MichaelGoodale: I've changed the `measure` to measure the drawn node (drawn by the `draw-node` callback) instead of just the Typst content.


MichaelGoodale:
Makes sense to measure the whole node! I had forgotten to remove that parameter too, thanks!

Issue #913, opened 2025-07-16
Multi-level style root


LordBaryhobal:
## Context
Styling is already quite flexible thanks to `styles.resolve`
However, the `root` argument only allows for a single level of styling
I think package developers would benefit from more levels, allowing at minimum to namespace all styles for their package into a single root
Furthermore, this could allow for easier styling of complex, hierarchical element

## Example
```typst
#{
ctx.styles = (
  circle: (...),
  rect: (...),
  ...,
  my-pkg: (
    element: (
      padding: 3pt,
      stroke: auto,
      fill: auto
    ),
    container: (
      stroke: blue,
      radius: 3pt,
      sub-element: (
        fill: red
      )
    )
    ...
  )
)
}
```

```typst
#styles.resolve(
  ctx.styles,
  style,
  root: ("my-pkg", "element"),
  base: default-element-style
)
```

johannes-wolf:
Closed by #914.

PR #914, opened 2025-07-18
styles: Support multi-level style root

LordBaryhobal:
This is a simple implementation for #913 adding support for an `array` value for the `root` of `styles.resolve`, allowing nested style resolving


johannes-wolf:
Thank you! Looks good.

PR #937, opened 2025-08-24
docs: fix link from merge-path to hide

NicoWeio:
The `@@` was quite striking; it was the only occurrence in the codebase.
Note that I did not yet set up a full-fledged environment (and I don't immediately see where and how the docs are built), so this fix is _untested_.

While looking at links throughout the docs, I also noticed that CeTZ functions are often unstyled, but sometimes `monospace`. Should I open another PR to move it all to monospace?

### DIFF
--- a/src/draw/shapes.typ
+++ b/src/draw/shapes.typ
@@ -1727,7 +1727,7 @@
 /// })
 /// ```
 ///
-/// Elements hidden via @@hide() are ignored.
+/// Elements hidden via [hide](../grouping/hide) are ignored.
 ///
 /// ## Anchors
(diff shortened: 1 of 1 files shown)


johannes-wolf (approved):
ty.

johannes-wolf:
> The `@@` was quite striking; it was the only occurrence in the codebase. Note that I did not yet set up a full-fledged environment (and I don't immediately see where and how the docs are built), so this fix is _untested_.
> 
> While looking at links throughout the docs, I also noticed that CeTZ functions are often unstyled, but sometimes `monospace`. Should I open another PR to move it all to monospace?

That would be awesome! They should be styled `monospace`.

Issue #808, opened 2025-02-02
Add near and far direction to the angle API


FalkDresden:
Prior to release 0.3.2 the angle API used to draw near angles (<= 180deg) per default, drawing far angles (> 180 deg) only when `inner` was set to `false`. This behavior was changed by #729 for reasons mentioned in #712.

While this solves the problem for some cases, it creates new problems elsewhere. Because now if you want to explicitly draw near/far angles, you first have to manually calculate whether to use the direction `"cw"` or `"ccw"`.

For this reason I propose to add two more directions:

- `near`: Behaves exactly like `inner: true` did in version 0.3.1
- `far`: Behaves exactly like `inner: false` did in version 0.3.1

PR #945, opened 2025-08-28
add near and far directions to angle api

FulcrumDev:
Adds the `near` and `far` directions to the angle function as described in #808.


johannes-wolf (approved):
LGTM. Thank you!

PR #950, opened 2025-08-30
fix: don't evaluate callback twice in get-ctx

Jollywatt:
Just a small thing I spotted. This doesn't affect cetz, but does affect the number of times sampled values appear in an ide.


johannes-wolf (approved):
👍

PR #985, opened 2025-11-07
test: Update tests to latest `tytanic` spec

dybucc:
`tytanic` reports nested tests as being deprecated, and eventually ignored in later versions. This PR simply pushes the changes performed by `tytanic`'s own `util migrate --confirm` subcommands and flag.

From skimming the diff, changes only include adding a more fine-grained directory structure and `.gitignore` files to both the `out/` and `dif/` subdirectories of each test, as those contain the resulting output of the test and the diff with the file in `ref/`, respectively.

While running tests, I also noticed the `tree` test failed. The circles surrounding the areas of the numbers were completely out of place, and it doesn't seem like the test in question has had any major changes in recent commits, so it may be that Typst's internals have modified something in the layout engine of more recent versions (seeing how `tytanic` doesn't seem to support Typst compiler versions beyond those of specific `tytanic` versions [[1]].) It seems unrelated to CeTZ, but I can't say for sure. What I can say is that the latest commit that changed the reference image for that specific test didn't change the `test.typ` file. Simply from observation alone, the resulting reference seems unmodified, but it could be there were some minor positioning changes. Either way, I've tried solving the test, and getting it to look half-decent again, but I've not opened a PR just yet because I'd require also updating the reference with `tytanic`'s corresponding command, and that's just pushing my luck with someone else's project.

[1]: https://typst-community.github.io/tytanic/reference/compat.html


johannes-wolf:
There is a `.gitignore` file in the `/tests/` directory that handles all the tytanic related files; no need to add it to every test case.

The default transformation matrix changed to an identity which might cause the problem.
Befor that change, some code had to flip the y-coordinate.

dybucc:
Done. My bad on the `tree` test; I don't know enough about the codebase yet to try to understand how to fix it.

Now `tytanic` shouldn't complain about nested tests anymore, and the additional `.gitignore` files have been removed.

PR #996, opened 2025-12-07
refactor: Rework docstring parser

dybucc:
The extractor needed some fine tuning to only pick up top level docstrings instead of (possibly wrong) locally scoped docstrings outside the reach of library users.

The function signature parser should now be a tad bit more efficient. It performs no operations on the whole array that gets passed beyond those strictly required by however as many elements (lines) are part of the function signature. Prior to this, there were a bunch of `enumerate` and `join` calls that wouldn't exactly be efficient for, say, arrays containing thousands of lines for some of the source code.

Further work will continue in the function parameter parsing, by possibly modifying the parameters of the parser itself, such that without modifying the resulting `typst query` output, we avoid performing two passes through the argument list; one in the initial function signature parsing, and another one in the parameter list parsing.

Once work on the function signature is done, the next step will be to fix the actual docstring parser, such that it picks up on newlines in function parameter documentation. An example of a docstring that I expect the parser to work through nicely is given in #986. Only after this is done, will I try to move on to seeing what can be done with the type syntax incompatibilities between the manual and the web documentation.

I also wanted to ask whether it's a good idea to be running the Python script for HTML generation directly and not through an isolated environment, by possibly using the nice integration `just` has with `uv` for script recipes [[1]].

[1]: https://just.systems/man/en/python-recipes-with-uv.html

### DIFF
--- a/docs/typlodocus/extractor.typ
+++ b/docs/typlodocus/extractor.typ
@@ -18,13 +18,12 @@
 
     if line.starts-with("///") {
       in-comment = true
-      current-comment += line.slice(3).trim() + "\n"
+      current-comment += line.slice(3) + "\n"
     } else if in-comment {
       if line.starts-with(regex(`#let\s+`.text)) {
-        let function = parser.parse-function-signature(lines.slice(i))
         comments.push((
-          comment: parser.parse-docstring(current-comment),
-          signature: function,
+          comment: parser.parse-docstring-alt(current-comment),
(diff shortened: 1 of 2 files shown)


johannes-wolf:
Hm, I have no opinion on this. But I would like to keep tools separate – the script should work without `just`. I guess expecting a working `python` executable on the host machine is fine, tbh.

PR #1034, opened 2026-02-22
Add matrix.transform-rotate-aer (azimuth/elevation/roll) helper

RaulDurand:
This PR adds a convenience rotation constructor to `matrix`:

`transform-rotate-aer(azimuth, elevation, roll: 0deg)`

## Behavior
- Builds a 4x4 rotation matrix from azimuth/elevation with optional roll.
- Uses the viewing convention:
  - `z` points up
  - `x` points toward the viewer
  - aligned with the common azimuth/elevation camera interpretation used by plotting tools (including matplotlib).
- Optional roll around the current viewing axis after azimuth/elevation.

## Implementation
- Added internal helper:
  - `_rotate-axis-angle(ax, ay, az, angle)` (axis-angle rotation matrix using the Rodrigues' formula)
- Added public function:
  - `transform-rotate-aer(azimuth, elevation, roll: 0deg)`

### DIFF
--- a/src/matrix.typ
+++ b/src/matrix.typ
@@ -180,7 +180,7 @@
   let rotate-azimuth = transform-rotate-z(-90deg - azimuth)
   let (ax, ay, az) = (-sin(azimuth), cos(azimuth), 0)
   let rotate-elevation = _rotate-axis-angle(ax, ay, az, elevation)
-  let base = mul-mat(ident(4), rotate-z-up, rotate-azimuth, rotate-elevation)
+  let base = mul-mat(rotate-z-up, rotate-azimuth, rotate-elevation)
 
   if roll == 0deg {
     return base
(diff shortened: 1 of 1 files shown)


src/matrix.typ
```diff
@@ -157,6 +157,41 @@
...
+/// - azimuth (angle): Rotation around z.
+/// - elevation (angle): Tilt above the xy plane.
+/// - roll (angle): Rotation around the current viewing axis.
+/// -> matrix
+#let transform-rotate-aer(azimuth, elevation, roll: 0deg) = {
+  let rotate-z-up = transform-rotate-x(-90deg)
+  let rotate-azimuth = transform-rotate-z(-90deg - azimuth)
+  let (ax, ay, az) = (-sin(azimuth), cos(azimuth), 0)
+  let rotate-elevation = _rotate-axis-angle(ax, ay, az, elevation)
+  let base = mul-mat(ident(4), rotate-z-up, rotate-azimuth, rotate-elevation)
```
johannes-wolf: Multiplying `ident` does nothing so it should be removed.

johannes-wolf (changes requested):
Nitpick, otherwise fine.

src/matrix.typ
RaulDurand: Indeed! I fixed it.

johannes-wolf (approved):

PR #1100, opened 2026-05-12
Fix broken assertion in `matrix.diag`

96-LB:
The `matrix.diag` function is currently unusable because of a couple typos in the assertion. It compares against an array instead of a dictionary (causing it to always fail the assertion), and there's a typo in the named argument (causing the assertion itself to error).

The line could alternatively be written as `assert(diag.named().len() == 0, ...` for consistency with the previous line, but I opted to edit it as minimally as possible.


johannes-wolf:
Oops :( My bad – seems I never tested it.

PR #1113, opened 2026-05-28
matrix: move transform functions to bottom of file

sijow:
Currently trying to use the function `transform-rotate-aer` fails with `unknown variable: mul-mat` because it's defined in the same file but before `mul-mat`. This PR moves the transform functions to the bottom of the file to let them use other matrix functions.

This also adds a test for `transform-rotate-aer`.


johannes-wolf (approved):
lgtm.

sijow:
I've pushed another commit to fix the typos in two rotation transforms.

PR #1115, opened 2026-06-26
docs: Update coordinate length handling

TheJanzap:
I added the information that the length of coordinates can be changed with `canvas(length: )` that was already present on "Canvas" and added it to "Coordinate Systems" for better visibility.

Additionally, I removed the mention that the parameter also takes ratios, due to the Canvas API docs mentioning that this behaviour was deprecated


johannes-wolf (approved):
Thank you!

Issue #1124, opened 2026-07-29
Harpoon stealth is not pointy.


iago-lito:
```typst
#import "@preview/cetz:0.5.2"
#cetz.canvas({
  import cetz.draw: *
  let s = (symbol: "stealth", stroke: red, fill: blue)
  line((0, .3), (1, .3), mark: s)
  line((0, 0), (1, 0), mark: (harpoon: true, ..s))
})
```

<img width="322" height="181" alt="Image" src="https://github.com/user-attachments/assets/1f2aa36f-5da2-4509-a969-b1c5cc8869f8" />

It looks like the marker stroke line becomes.. unclosed at the tip?

iago-lito:
I guess this is just an unfortunate consequence of the following "pointiness unstability".. ?
```typst
line((0, 0), (1, 0), (0, 0.6), close: true, stroke: blue)
line((0, -1), (1, -1), (0, 0.5 - 1), close: true, stroke: red)
```

<img width="330" height="502" alt="Image" src="https://github.com/user-attachments/assets/f96f0d5a-3c84-41cd-94c9-193f43d2f728" />

iago-lito:
Hm, making progress ^ ^ https://typst.app/docs/reference/visualize/stroke#constructor-miter-limit.

iago-lito:
Workaround in the meantime: set `mark: (stroke: (miter-limit: 100))` or whatever large number.

PR #1125, opened 2026-07-29
Bump default miter-limit for stealth marker.

iago-lito:
This is a hacky first shot at #1124.
I am not liking it because it overrides user-defined `miter-limit` if they chose an explicit one for some reason.
Can someone provide guidance how to better raise this default value for *arrows markers* specifically so they remain pointy by default as most likely expected?
(My guess is that very few users want their arrow beveled, so we can pick a rather high default value.)

### DIFF
--- a/src/mark-shapes.typ
+++ b/src/mark-shapes.typ
@@ -157,7 +157,7 @@
       ((0,0), (l, w / 2), (l - i, 0), (l, -w / 2))
     }
 
-    fast-line(..pts, stroke: style.stroke, fill: style.fill, close: true)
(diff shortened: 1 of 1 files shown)


johannes-wolf:
Styling specific mark shapes using cetz' styling system is currently not possible. But should be easy to add. I will take a look. Then we can give a big miter limit to stealth markers.

iago-lito:
Cool, thank you @johannes-wolf I suppose not only `stealth` will benefit from this but [any _pointy_ marker type](https://cetz-package.github.io/docs/basics/marks).

PR #1091, opened 2026-05-01
Implement path boolean operations

hooyuser:
## Description
This PR introduces the `path-bool` function to support boolean operations on paths. The function accepts two cetz bodies, converts them into path-like objects consisting strictly of closed subpaths, applies the requested boolean operation (`union`, `intersection`, `difference`, or `xor`), and returns a new path-like object containing only the resulting closed subpaths.

The underlying boolean operation engine is powered by [linesweeper](https://docs.rs/linesweeper/latest/linesweeper/), a MIT/Apache-2.0 dual-licensed Rust crate (notably used by the [Graphite](https://github.com/GraphiteEditor/Graphite) project for their path boolean operations).

To bridge cetz's 3D paths and linesweeper's 2D paths, this PR introduces WirePath as an intermediary 2D path structure. The complete data flow pipeline is as follows:

> CeTZ Body → CeTZ Path3D → WirePath2D → BezPath → [Boolean Operation] → BezPath → WirePath2D → CeTZ Path3D → CeTZ Drawable

To keep this PR concise and easier to review, I have only included a curated subset of the CeTZ and Rust test cases in this initial push.

## Next Steps
The next goal is to further leverage `linesweeper` to implement path clipping (using a closed path to clip other open or closed paths). Ultimately, this will serve as the foundation for clipping elements within a group, addressing the issue #618.

Issue #618, opened 2024-05-31
Implement clipping region for groups


hooyuser:
Hi, I was wondering if there is a possibility to implement clipping regions for groups, similar to how it's handled in TikZ. In TikZ, we can define a clipping path within a scope to constrain the visible area of contained elements, as demonstrated below:
```
\begin{scope} 
    \clip (-2, 0) rectangle (2, 2);     
\end{scope}
```
Would it be possible to introduce a similar feature for groups in CeTZ, allowing users to define clipping regions that limit rendering to specified boundaries? This feature would be incredibly helpful for complex diagrams where elements outside a certain area need to be hidden.

johannes-wolf:
This is planned, but doing this for non-rectangular regions is not trivial.

johannes-wolf:
I guess you mean path-clipping (boolean operations on paths)?

hooyuser:
Thank you for your quick response!

Yes, I was referring to path-clipping similar to what is shown in the examples on [this TikZ page](https://tikz.org/examples/chapter-07-filling-and-clipping/). However, even just implementing clipping for rectangular regions will help a lot for my current needs.

Could you please elaborate on whether it's possible to set up a rectangular clipping region for groups with the current version of CeTZ? Any guidance or workarounds would also be greatly appreciated.

johannes-wolf:
With the current version of cetz it is possible but not implemented. You have to implement some form of path clipping. There is a closed PR #535 that implements this by using a wasm library.

hooyuser:
Given the complexities of general clipping, I'm curious to learn how to implement rectangular clipping first. Could you provide some guidance or hints?

It’s a pity about the licensing issues that led to the closure of PR #535, especially considering the effort that must have gone into it. While I haven’t looked into this extensively, a preliminary search suggested that there may be some other suitable path-clipping libraries like [Clipper2](https://github.com/AngusJohnson/Clipper2) compatible with LGPL. I thought this might be worth mentioning for future considerations.

ramkumarkb:
Echoling @hooyuser ' s comment - was [Clipper2](https://github.com/AngusJohnson/Clipper2) considered, especially it has friendly license and a WASM port too.

johannes-wolf:
> Echoling @hooyuser ' s comment - was [Clipper2](https://github.com/AngusJohnson/Clipper2) considered, especially it has friendly license and a WASM port too.

[...]

PR #1131, opened 2026-08-26
Implement path clipping with exact Bézier curves

hooyuser:
This PR builds on top of the path boolean work from #1091 and is the next step toward addressing #618.

It adds a new `clip` function to clip path drawables against a closed clipping region:

```typst
clip(
  clipping-region,
  body,
  mode: "inside",
  ...
)
```
Depending on `mode` (`"inside"` or `"outside"`), it keeps only the geometry inside or outside the `clipping-region`. The `body` parameter takes the objects you want to clip, and each path drawable inside it is clipped independently against `clipping-region`. Note that unlike paths in `clipping-region`, paths in `body` may contain both open and closed subpaths.

Take `mode: "inside"` as an example. Under the hood, the `stroke` and `fill` of each path drawable in `body` are handled separately:

- If its `stroke` is not `none`: subpaths (open or closed) are trimmed to keep only the portions inside the clipping region.
- If its `fill` is not `none`:
  - Its closed subpaths together define the filled area according to `body-fill-rule`. A boolean intersection between that area and the `clipping-region` produces a newly generated, fill-only path.
  - Its open subpaths do not participate in the Boolean operation.

`mode: "outside"` follows the same split: it retains the outside portions of stroked paths and uses a boolean difference for filled areas.

To avoid repeating logic from #1091, I also reorganized the relevant Typst and Rust code into shared Typst/Rust modules. This is purely an internal cleanup and doesn't change the behavior of the existing `boolean` function.

---

### Example: string diagram of snake equation

Someone on Discord asked how to draw a string diagram and came up with a solution using boolean operations ([discussion link](https://discord.com/channels/1054443721975922748/1260973060585164810/1533122580519850355)). With `clip`, we now have a more straightforward way to achieve this kind of effect.

```typst

#cetz.canvas({
  import cetz.draw: *

  let padded-anchor(name, position) = cetz.draw.group(name: name, {
    import cetz.draw: *

    let padding = 0.3
    anchor("default", position)
    anchor("north", (position.at(0), position.at(1) + padding))
    anchor("south", (position.at(0), position.at(1) - padding))
    anchor("east", (position.at(0) + padding, position.at(1)))
    anchor("west", (position.at(0) - padding, position.at(1)))
  })

  let path = {
    line("F_in", (4, 2))
    bezier((4, 2), "epsilon.east", (4, 3))
    bezier("epsilon.west", (2, 2), (2, 3))
    bezier((2, 2), "eta.east", (2, 1))
    bezier("eta.west", (0, 2), (0, 1))
    line((0, 2), "F_out")
  }

  on-layer(10, {
    circle((1, 1), radius: 3mm, name: "eta", fill: white)
    content((), $eta$)

    circle((3, 3), radius: 3mm, name: "epsilon", fill: white)
    content((), $epsilon$)

    padded-anchor("F_in", (4, 0))
    content("F_in.south", $F$)

    padded-anchor("F_out", (0, 4))
    content("F_out.north", $F$)

    path
  })

  let background(fill) = {
    rect(
      (-0.5, 0),
      (4.5, 4),
      stroke: none,
      fill: fill,
    )
  }

  let aqua-region = {
    merge-path(
      {
        path
        line("F_out", (-0.5, 4), (-0.5, 0))
      },
      close: true,
    )
  }

  on-layer(0, {
    background(lime)
    clip(aqua-region, background(aqua))
  })
})
```
<p align="center">
  cetz
</p>


src/draw/shapes.typ
```diff
@@ -2173,7 +2173,7 @@
       if cmd in ("z", "Z") {
         assert.eq(args.len(), 0)
         if current != () {
-          paths.push(path-util.make-subpath(origin, current, closed: cmd == "z"))
+          paths.push(path-util.make-subpath(origin, current, closed: true))
```
johannes-wolf, line 2176: 👍

johannes-wolf (commented):
LGTM! Thank you a lot!

PR #1132, opened 2026-08-28
Fix equation anchors when the baseline is outside the bounding box

### LINKED ISSUE #1123: Anchors can't optically center a glyph that doesn't reach cap-height
All `content()` anchors derive from font metrics, so a short symbol (`+`, `−`, `×`, `⋅`)
can't be centered in a shape.

```typ
#import "@preview/cetz:0.5.2": canvas, draw
#import draw: circle, content, line
#set page(width: auto, height: auto, margin: 10pt, fill: white)

#canvas({
  for (idx, a) in ("center", "mid", "base").enumerate() {
    let x = idx * 2.2
    circle((x, 0), radius: 0.6, stroke: 0.6pt)
    line((x - 0.6, 0), (x + 0.6, 0), stroke: (paint: green, thickness: 0.4pt))
    content((x, 0), text(size: 20pt)[+], anchor: a)
  }
})
```

Image

Green marks each circle's center. `center` uses the full line box, including ascent and
descent the glyph never occupies. `mid` is `lerp(cap-height-top, baseline, 0.5)`, and a
`+` doesn't reach cap-height either, so it lands *lower* than `center`. `base`
overshoots.

The only workaround is a hand-tuned `text(baseline: ...)` per glyph and per size.

**Suggestion:** an anchor from the content's ink bounds — `anchor: "ink-center"` or a
`bounds: "ink"` option. cetz already measures `bottom-edge: "bounds"` for
`baseline-offset`, so this may just need the matching `top-edge: "bounds"`.

Related: #683 added `mid`/`base`; #794 is a different font-metric mismatch.

cetz 0.5.2, Typst 0.15.0

hooyuser:
For equations such as `$-$`, the baseline can lie below the equation's bounding box. In other cases, the baseline can lie above the equation's bounding box.

This PR handles both edge cases and adds tests for them.


johannes-wolf (approved):
LGTM.

